PR #8, opened 2017-04-09
Fix issue with recent deals

anthonyp:
Deals that have already been synced in the last 30 days use a different endpoint for a "delta" sync. This endpoint returns the results in a slightly different format, but as it stood, the parser did not account for it. This pull request addresses the issue.


mdelaurentis:
Thanks for the contribution. Merging now.

PR #13, opened 2017-04-21
Store current stream and offset in state

mdelaurentis:


tap_hubspot/__init__.py
```diff
@@ -459,29 +472,48 @@ def sync_owners():
...
+            lambda x: x.name != state[StateFields.this_stream], streams))
+
+
 def do_sync():
     LOGGER.info("Starting sync")
-    for name, func in STREAMS:
-        LOGGER.info('Syncing %s', name)
-        func()
+    streams = get_streams_to_sync(STREAMS, STATE)
+    LOGGER.info('I will sync these streams: %s',
```
KAllan357: Should the Tap be logging in the first person?

PR #22, opened 2017-06-09
Added a number of important attributes to the HubSpot tap.

jdwyah:
These properties only show up in some events and are not listed in the docs, but are quite important.

### DIFF
--- a/tap_hubspot/schemas/email_events.json
+++ b/tap_hubspot/schemas/email_events.json
@@ -34,9 +34,24 @@
       "type": ["null", "string"],
       "format": "date-time"
     },
+    "deviceType": {
+      "type": ["null", "string"]
+    },
+    "duration": {
+      "type": ["null", "integer"]
+    },
     "emailCampaignId": {
       "type": ["null", "integer"]
     },
+    "emailCampaignGroupId": {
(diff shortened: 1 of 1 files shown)


jdwyah:
@jkebinger anything else I should add?

briansloane:
@jdwyah Thanks for your contribution. These are fields that we have been unable to test within our HubSpot sandbox. Have you been able to successfully run the Tap in a development environment to confirm these fields are properly output and data typed as per the schema modfications?

mxie:
@briansloane Not surprised that `emailCampaignGroupId` is not there for you - it only shows up when emails are sent in customer local time zones. Tried this out in development mode and everything seems to output fine.

briansloane:
Thanks, @mxie! We'll merge this and update Stitch with the latest version.

jdwyah:
awesome!  and @briansloane fyi @mxie will have a new engagements one for you shortly, so if it's a process to update you may want to wait for that one too. Seems likely we could have that by EOD.

briansloane:
@jdwyah Thanks for the heads up. I see the other PR now. We'll handle them separately since the other one is more than a few updates to the schema.

PR #41, opened 2017-08-16
add “disposition” to engagements

jdwyah:
This field exists on phone calls and tracks whether the call was successful.
Its actual value is a guid, but it can be deciphered.


karstendick:
@jdwyah Thanks for the patch! We've merged it in here: https://github.com/singer-io/tap-hubspot/pull/43

PR #53, opened 2017-10-23
 Enables use of a HubSpot HAPIKEY for authentication

jaeger401:
This PR adds one-step configuration for API key usage, intended to ease the development process. It also updates README to explain API key usage and adds a sample config file to the repo. (This partially addresses issue #9.)

The verbiage in the README was drawn from similar language in the [singer-io/target-csv](https://github.com/singer-io/target-csv/blob/2e76485b2d574594387a945aeb7d6e7df4c43645/README.md) repo.

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -51,6 +51,7 @@ CONFIG = {
     "client_secret": None,
     "refresh_token": None,
     "start_date": None,
+    "hapikey": None
 }
 
 
@@ -231,7 +232,7 @@ def parse_source_from_url(url):
 @utils.ratelimit(9, 1)
 def request(url, params={}):
 
-    hapikey = os.getenv("HUBSPOT_HAPIKEY")
+    hapikey = CONFIG['hapikey']
     if hapikey is None:
         if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
             acquire_access_token_from_refresh_token()
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -229,13 +229,17 @@ def parse_source_from_url(url):
                       on_giveup=on_giveup,
                       factor=2)
 @utils.ratelimit(9, 1)
-def request(url, params=None):
+def request(url, params={}):
 
-    if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
-        acquire_access_token_from_refresh_token()
+    hapikey = os.getenv("HUBSPOT_HAPIKEY")
+    if hapikey is None:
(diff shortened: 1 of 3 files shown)


jaeger401:
Hey @mdelaurentis, I pulled out the environment variable and swapped in reading from `config.json` as requested. 

Apologies for the mixed config source there -- initially, it seemed confusing to have the possibility of two different auth sources in that config. I updated it so that the API key has priority if present, mirroring how the env variable was handled. Feel free to object to that reasoning; it's easy enough to give OAuth priority instead. Thanks.

tap_hubspot/__init__.py
```diff
@@ -229,13 +230,17 @@ def parse_source_from_url(url):
                       on_giveup=on_giveup,
                       factor=2)
 @utils.ratelimit(9, 1)
-def request(url, params=None):
+def request(url, params={}):
```
iterati: This is a common "gotcha" in Python, but default mutable arguments are bad.

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

tap_hubspot/__init__.py
```diff
@@ -229,13 +230,17 @@ def parse_source_from_url(url):
...
 
-    if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
-        acquire_access_token_from_refresh_token()
+    hapikey = CONFIG['hapikey']
+    if hapikey is None:
+        if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
+            acquire_access_token_from_refresh_token()
+        headers = {'Authorization': 'Bearer {}'.format(CONFIG['access_token'])}
+    else:
+        params = {**params, **{'hapikey': hapikey}}
```
iterati: Why that vs `params['hapikey'] = hapikey`? This is very unidomatic and confusing.

tap_hubspot/__init__.py
jaeger401: Thanks -- I've been away from Python for long enough that I'd forgotten about that.

tap_hubspot/__init__.py
jaeger401: You are correct. I've made that amendment. 

My initial implementation had more than just the one `hapikey` key in the dictionary, hence the code to merge with `params`. This code was the product of a thought process that started with "how do I do this in Python 3?" and  spiraled to a confusing place.

iterati: No worries. Merging dicts can be done with `.update`:

```
In [1]: params = {"a": 1, "b": 2}
   ...: params.update({"c": 3, "d": 4})
   ...: params
   ...:
Out[1]: {'a': 1, 'b': 2, 'c': 3, 'd': 4}
```

PR #55, opened 2017-11-30
fix contacts link in README

iterati:
@erinkcochran87 I redid this against the latest master. My gitfu isn't where it needs to be to figure out why there were so many changes required for your branch or why I was getting conflicts when rebasing. Does this look good?


erinkcochran87:
@iterati Looks superb. Thanks for fixing it!

luandy64 (approved):

PR #58, opened 2017-12-18
Syncs contacts based on `versionTimestamp`

jaeger401:
This pull request addresses a long-standing issue with list memberships in which memberships are dependent on whether or not a contact has been updated, and thus new dynamic list memberships are frequently not captured.

HubSpot has recently exposed a new pair of fields (`version` & `versionTimestamp`) which are returned when the `includeVersion` query parameter is set to `true` when calling the `contacts` APIs. These new fields track the internal versions of the contacts, which includes updates due to new dynamic list memberships. Updates to other contact properties will also trigger a new `versionTimestamp`, so no prior functionality is lost. Thanks to @jaredmiwilliams for the change.

I've tested a number of scenarios against our dev HubSpot account to confirm functionality, and everything works as expected.

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -307,11 +307,12 @@ def _sync_contact_vids(catalog, vids, schema, bumble_bee):
 
 default_contact_params = {
     'showListMemberships': True,
+    'includeVersion': True,
     'count': 100,
 }
 
 def sync_contacts(STATE, catalog):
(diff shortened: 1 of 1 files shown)


jdwyah:
@jaredmiwilliams #fyi & thankyou

nick-mccoy:
Thanks for the contribution @jaeger401!  I'm going to run a few tests to validate your changes and then I'll update with results.

PR #60, opened 2018-01-09
Add options to help minimize api requests during company sync

mikeclarke:
We have a ton of objects in Hubspot, and we have to be careful to manage the 40,000 requests per day daily limit. The additional individual `company_detail` and `contacts_by_company` calls explode our request count, though the `companies_all` endpoint retrieves 250 rows in a batch.

I'm not familiar with the best way to configure this in the singer world, but this patch adds support for a new `metadata` field:

```
{
  "stream": "companies",
  "tap_stream_id": "companies",
  "metadata": {
    "additional_properties": [
      "custom_company_property_1"
    ],
    "sync_contacts_by_company": false
  },
  "schema": { ... }
```

The `additional_properties` list implies that the detail requests are unnecessary, since the original response contains the fields already.

This patch also adds a `sync_contacts_by_company` flag to disable the generation of the contact-company records.

I wasn't sure of the best way to test this given the tests already in place, but I can add some if required. This change should be backwards-compatible with the existing behavior.

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -380,12 +380,16 @@ def sync_companies(STATE, catalog):
     LOGGER.info("sync_companies from %s", start)
     schema = load_schema('companies')
     singer.write_schema("companies", schema, ["companyId"], catalog.get('stream_alias'))
+    metadata = catalog.get('metadata', {})
 
     url = get_url("companies_all")
     max_bk_value = start
 
     with bumble_bee:
-        for row in gen_request(STATE, 'companies', url, default_company_params, 'companies', 'has-more', ['offset'], ['offset']):
+        params = default_company_params
+        if metadata.get('additional_properties'):
+            params['properties'] = params['properties'] + metadata['additional_properties']
+        for row in gen_request(STATE, 'companies', url, params, 'companies', 'has-more', ['offset'], ['offset']):
             row_properties = row['properties']
             modified_time = None
             if 'hs_lastmodifieddate' in row_properties:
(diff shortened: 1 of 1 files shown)


cmerrick:
Hi @mikeclarke, thanks for your contribution!

cmerrick:
You did it @mikeclarke!

KAllan357:
Hi @mikeclarke thanks for the contribution. Apologies for the delayed response.

I want to make sure I understand this change so let's break it down into its distinct parts:

First, you'd like a way to retrieve additional fields from the initial `/companies` request by adding some user-defined values into the request's query params. This is currently implemented in your PR via the metadata property `additional_properties`. When this property is present, the tap will refrain from calling the subsequent `/companies/{ID}` calls.

Second, you'd like to be able to decide whether or not the tap should sync Companies by Contacts via the calls to the `/companies/v2/companies/{company_id}/vids` endpoint. This allows you to preserve your API limit because the data is not necessary.

Does that all sound correct?

mikeclarke:
That's correct - for the first part, the tradeoff is having to be explicit about the properties you'd want ahead of time and saving API calls vs. pulling all of the data and consuming additional API calls. Seems like both use cases are handy to have, but maybe there's a way to get the best of both worlds by refactoring to detect available properties and including them by default in the `/companies` request. I'm not sure if that's 100% backwards compatible or if there's additional data that is only available in the company detail calls.

KAllan357:
I agree - both use cases have value. I had a conversation with my coworkers and I think we came up with a solution.

I'd like to split this PR / issue into the two chunks:

1. Enable the tap to retrieve extra properties for the Company object without using the company_details endpoint because it could max out Hubspot's API limits.

There are some extra considerations we need to make on the Stitch side, so I think this task should be handled by us.

2. Implement the "field selection" [TODO here](https://github.com/singer-io/tap-hubspot/blob/master/tap_hubspot/__init__.py#L357). Doing this plus a little extra work will all us to stop doing executing those calls if we don't select the `contacts_by_company` table.

I believe this task can be completed entirely within the tap - and if you are interested in tackling it I'd happily merge. The broad strokes to the work would be:

* Add a new Stream for `contacts_by_company` [here](https://github.com/singer-io/tap-hubspot/blob/master/tap_hubspot/__init__.py#L707) 
* Using a similar pattern as [tap-jira](https://github.com/singer-io/tap-jira/blob/master/tap_jira/streams.py#L286) we want to make sure that if someone selects `contacts_by_company`, that its required that they have Companies selected. In other words, `contacts_by_company` depends on `Companies`.
* Replace the metadata check in [your PR](https://github.com/mikeclarke/tap-hubspot/blob/07e7d6e0f88f03eccf62aae2aaea3b56fa7f8849/tap_hubspot/__init__.py#L418) with a check to see if `contacts_by_company` is selected. (this step requires some thought because each `sync_` function only receives the catalog it knows about instead of all of them)

If you don't have the bandwidth to take on this second task, I've also gone ahead and added it to our backlog so we can take care of it.

nick-mccoy:
Part 2 completed in #63

PR #62, opened 2018-01-24
emit bookmark_properties for schema and time_extracted for records

ccapurso:
singer-python version 5.0.0 introduced time_extracted for the `RecordMessage` and bookmark_properties for the `SchemaMessage`. We are now going to support these in tap-hubspot.


ccapurso:
Yes, I was able to get this to pass in tap-tester. I also functionally tested.

PR #71, opened 2018-05-08
Remove form-submissions that are timestamped less than the bookmark

KAllan357:


KAllan357:
This will cause some weird behaviors because we don't bookmark / filter nested data. The true fix here is to add a new stream for this data and filter it.

Issue #77, opened 2018-10-08
Ensure deals associations are being extracted


stephlocke:
Further to a convo with Stitch, the Associations array on the Deals API is being lost. They requested I raise it as an Issue here.

This is the *only* connection to companies and contacts making is very important to have for downstream reporting. 

## Things to do 

- [ ] Validate associations are correctly being retrieved in this tap
- [ ] Add unit tests to verify automatically
- [ ] Fix code if required

KAllan357:
Thanks for the issue @stephlocke. Would you be open to making a PR for this fix? 

I think this could be achieved by adding the `includeAssociations` param to the [parameter map here](https://github.com/singer-io/tap-hubspot/blob/master/tap_hubspot/__init__.py#L441). It should inspect the `mdata` map to determine whether the field is selected and if so the value for `includeAssociations` should be `True`

stephlocke:
I can take a look but it's pretty unlikely I can contribute - python isn't my key language and I'm coming to these repos from stitch versus implementing singer

BogdanCojocar:
Hi, I am trying to submit a PR for this issue, but I get 403. Can I be authorised to write please?

stephlocke:
@BogdanCojocar it might have been an intermittent issue - I just did a sample PR and it worked alright https://github.com/singer-io/tap-hubspot/pull/78

timvisher:
@BogdanCojocar Thanks for working on the issue. It must be a temporary blip. There's nothing on our end that should prevent that that I know of.

timvisher:
@BogdanCojocar Ah sorry I see that you did submit the PR. Thanks! We'll review it asap. :)

BogdanCojocar:
@timvisher Yeah, sorry, it was an issue on my side. Thanks!

dmosorast:
#79 has been merged, and I will be releasing version `2.2.4` with this change. @stephlocke Please close this issue if it fixes what you are seeing, or update it if there are further changes to be made. Thanks!

PR #79, opened 2018-10-29
add check for associations

BogdanCojocar:
PR for this issue https://github.com/singer-io/tap-hubspot/issues/77

Tested with the following stream:
```
{
  "streams": [
    {
      "tap_stream_id": "deals",
      "stream": "deals",
      "schema": {
        "type": ["null", "object"],
        "selected": "true",
        "additionalProperties": "false",
        "properties": {
          "portalId": {
            "type": [
              "null",
              "integer"
            ]
          },
          "dealId": {
            "type": [
              "null",
              "integer"
            ],
            "selected": "false"
          },
          "associations": {
            "type": ["null", "object"],
            "properties": {
              "associatedVids": {
                "type": ["null", "array"],
                "items": {
                  "type": ["null", "integer"]
                }
              },
              "associatedCompanyIds": {
                "type": ["null", "array"],
                "items": {
                  "type": ["null", "integer"]
                }
              },
              "associatedDealIds": {
                "type": ["null", "array"],
                "items": {
                  "type": ["null", "integer"]
                }
              }
            }
          }
        }
      },
    "metadata": [
        {
          "metadata": {
            "inclusion": "available",
            "table-key-properties": ["portalId"],
            "selected-by-default": true,
            "valid-replication-keys": ["portalId"],
            "schema-name": "deals",
            "selected": true
          },
          "breadcrumb": []
        },
        {
          "metadata": {
            "inclusion": "automatic",
            "selected": true
          },
          "breadcrumb": ["properties", "portalId"]
        },
        {
          "metadata": {
            "inclusion": "available",
            "selected-by-default": true,
            "selected": true
          },
          "breadcrumb": ["properties", "dealId"]
        },
        {
          "metadata": {
            "inclusion": "automatic",
            "selected": true
          },
          "breadcrumb": ["properties", "associations"]
        }
      ]
    }
  ]
}
```

I see that the unit tests that check the **sync** methods (in `test_bookmarks`, `test_offsets`) are either incomplete or broken so I have not added a new unit test. 

Please let me know if you need additional info.

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -439,10 +439,18 @@ def sync_deals(STATE, ctx):
     LOGGER.info("sync_deals from %s", start)
     most_recent_modified_time = start
     params = {'count': 250,
+              'includeAssociations': False,
               'properties' : []}
 
     schema = load_schema("deals")
     singer.write_schema("deals", schema, ["dealId"], [bookmark_key], catalog.get('stream_alias'))
+
+    # Check if we should  include associations
+    for key in mdata.keys():
+        if 'associations' in key:
+            assoc_mdata = mdata.get(key)
+            if (assoc_mdata.get('selected') and assoc_mdata.get('selected') == True):
+                params['includeAssociations'] = True
 
     # Append all the properties fields for deals to the request
     additional_properties = schema.get("properties").get("properties").get("properties")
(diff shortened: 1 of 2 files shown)


dmosorast:
Hi @BogdanCojocar Thanks for submitting this PR! Can you provide a gist of logs (with any sensitive information redacted) that show the tap is requesting the associations and that they are getting persisted with the messages with no fields logged as "removed"? My concern here is that the schema matches the expected output from a request with this parameter specified.

Also, I took a look at the tests that currently exist, and it does look like they are older integration tests that are mostly no-ops right now. Given that the linked issue is requesting a unit test, could you write a test that calls `sync_deals` directly and replaces the `gen_request` function with one that validates the parameters passed to this? That will help ensure that this doesn't get lost in a future refactoring. Thanks!

dmosorast:
@BogdanCojocar Great! Thank you for the gist, this looks good!

For the test, that is correct. Given the complexity of the code in this area, mocking `gen_request` and checking the parameters will help prevent regressions in future changes.

dmosorast:
Great! Thank you. This looks good to merge.

PR #85, opened 2019-02-11
N/A values in numeric fields

mirelagrigoras:
I have been trying to use tap-hubspot with target-postgres, but I had a problem with the integer columns. For fields that have the type 'number',  tap-huspot will two types for a single field ("number" and "string", in order to accept 'N/A' values : https://github.com/singer-io/tap-hubspot/blob/master/tap_hubspot/__init__.py#L120 ), which is not allowed in a relational database (a column must have a single data type). Generating 2 field types will make target-postgres (or any other sql target) to create 2 columns of different types and different names from the original field name, which will make the import of the hubspot data impossible, since a field should be mapped 
to only one column, not to two. 
So, my suggestion would be to process the rows as they are received from the APIs calls and replace any 'N/A' values with None, so that the type of a number field could be  
 `"type":[ "null", "number" ]`.

Let me know if I`m missing something or if I missunderstood anything regarding how tap-hubspot works
Also, if you have any another approach that would solve this problem, I would be more than pleased to hear it.


cmerrick:
Hi @mirelagrigoras, thanks for your contribution!

cmerrick:
You did it @mirelagrigoras!

mirelagrigoras:
@cmerrick Could you, please, take a look on this pull request?

PR #86, opened 2019-02-21
Postgres compatible tap-hubspot

mirelagrigoras:
Tap-hubspot, as it is now, generates a schema with 2 types for integer fields, which is not allowed in a Postgres database (a column must have a single data type) and will make the postgres target (or any other relational databases targets) to generate 2 columns with different types and different names. 
In this pull request I add a new argument that can be passed to the tap `--postgres_compatible` which will make the tap replace N/A values with None  and generate a schema that will contain `"type":[ "null", "number" ] ` for integer fields, not `"type":[ "null", "number" , "string"]`, as it happens in the current version. Otherwise, if the `--postgres_compatible` argument is not passed, the schema that is generated will remain the same as it is at the moment.

This is an alternative to https://github.com/singer-io/tap-hubspot/pull/85/

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -920,7 +920,7 @@ def do_sync(STATE, catalogs, postgres_compatible = False):
         singer.write_state(STATE)
 
         try:
-            STATE = make_postgres_comp_if_needed(stream, STATE, ctx, postgres_comp)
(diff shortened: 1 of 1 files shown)


paddie:
@mdimercurio thanks for this, I used it to solve the issue I had with the target-bigquery - which is also breaking on multiple types.

PR #94, opened 2019-07-22
Feature/configurable window size

luandy64:
Hard coding the `chunk_size` has seemed to work for the tap up until now. This PR renames `chunk_size` to `window_size` and makes them a configurable field via `config.json`

There's also some whitespace changes that happened.

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -38,11 +38,6 @@ class StateFields:
     offset = 'offset'
     this_stream = 'this_stream'
 
-CHUNK_SIZES = {
-    "email_events": 1000 * 60 * 60 * 24,
-    "subscription_changes": 1000 * 60 * 60 * 24,
-}
-
 BASE_URL = "https://api.hubapi.com"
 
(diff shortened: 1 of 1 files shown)


tap_hubspot/__init__.py
```diff
@@ -38,18 +38,15 @@ class StateFields:
...
-
 BASE_URL = "https://api.hubapi.com"
 
 CONTACTS_BY_COMPANY = "contacts_by_company"
 
 CONFIG = {
     "access_token": None,
     "token_expires": None,
+    "email_chunk_size": 1000 * 60 * 60 * 24,
+    "subscription_chunk_size": 1000 * 60 * 60 * 24,
```
dmosorast: Assuming this gets overridden when args are parsed? Seems a bit too magic, can you make it fall back at that point? Keeping it None here? With a constant like `DEFAULT_CHUNK_SIZE`?

luandy64: My understanding is that the args will upsert into this dictionary. 

I don't think None is the right move here. But I'll pull out the numbers as `DEFAULT_CHUNK_SIZE`

PR #100, opened 2019-09-05
Switched to using metadata in V2

luandy64:


tap_hubspot/__init__.py
```diff
@@ -864,9 +861,13 @@ def do_sync(STATE, catalogs):
...
 class Context(object):
     def __init__(self, catalog):
-        self.selected_stream_ids = set(
-            s.get('tap_stream_id') for s in catalog.get('streams')
-            if s.get('schema').get('selected'))
+        self.selected_stream_ids = set()
+
+        for stream in catalog.get('streams'):
+            for mdata in stream['metadata']:
+                if mdata['breadcrumb'] == [] and mdata['metadata']['selected'] == True:
```
dmosorast: Should probably be `mdata['metadata'].get('selected')`

tap_hubspot/__init__.py
```diff
@@ -864,9 +861,13 @@ def do_sync(STATE, catalogs):
...
+                    self.selected_stream_ids.add(stream.get('tap_stream_id'))
```
dmosorast: This should always be present, and it might be better to blow up rather than to have a `tap_stream_id` of None. Should be `stream['tap_stream_id']`

tap_hubspot/__init__.py
```diff
@@ -824,26 +825,22 @@ def get_streams_to_sync(streams, state):
...
-
     return selected_streams
 
-def do_sync(STATE, catalogs):
+def do_sync(STATE, catalog):
     # Clear out keys that are no longer used
     clean_state(STATE)
 
-    ctx = Context(catalogs)
+    ctx = Context(catalog)
```
dmosorast: I didn't see this in the other commit (for crossbeam's version). Doesn't look like an issue, but wanted to comment for a second look.

luandy64: We changed this during our first read through the code. But it kept it because it fit our mental model better. We didn't change it in the crossbeam version though

dmosorast: Sounds good to me!

PR #107, opened 2020-01-02
Move properties fields to top level

KAllan357:
# Description of change

Moves properties for certain streams (`deals`, `contacts`, `companies`) to the top level so they are selectable and also keep them nested under `properties`.

Moves versions of data to the top level and nests under `properties_versions`

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -162,7 +162,7 @@ def get_field_schema(field_type, extras=False):
 def parse_custom_schema(entity_name, data):
     # Prefix custom properties with "property"
     return {
-        "property_{}".format(field['name']): get_field_schema(
+        field['name']: get_field_schema(
             field['type'], entity_name != "contacts")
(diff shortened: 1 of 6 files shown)


tap_hubspot/__init__.py
```diff
@@ -836,6 +859,13 @@ def get_selected_streams(remaining_streams, ctx):
             selected_streams.append(stream)
     return selected_streams
 
+def has_selected_custom_field(mdata):
+    top_level_custom_props = [x for x in mdata if len(x) == 2 and 'property_' in x[1]]
+    for prop in top_level_custom_props:
+        if mdata.get(prop, {}).get('selected') == True:
+            return True
+    return False
+
```
dmosorast: Move this up to above `sync_deals`, just for order-s sake (even though Python doesn't care)

PR #124, opened 2020-09-03
Add hs_date_entered/exited fields from v3

luandy64:
# Description of change
The v3 CRM Deals API has `hs_date_entered_*` and `hs_date_exited_*` fields that the v1 Deals API does not have. This PR discovers and syncs those fields.

# Manual QA steps
 - Ran the tap
 
# Risks
 - Low because everything is an addition
 
# Rollback steps
 - revert this branch

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,13 @@ setup(name='tap-hubspot',
           'requests==2.20.0',
           'backoff==1.3.2',
           'requests_mock==1.3.0',
-          'nose'
       ],
+      extras_require= {
+          'dev': [
+              'pylint==2.5.3',
+              'nose==1.3.7',
+          ]
+      },
       entry_points='''
           [console_scripts]
           tap-hubspot=tap_hubspot:main
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -72,8 +72,8 @@ ENDPOINTS = {
     "companies_detail":     "/companies/v2/companies/{company_id}",
     "contacts_by_company":  "/companies/v2/companies/{company_id}/vids",
 
-    "deals_properties":     "/properties/v1/deals/properties",
-    "deals_all":            "/deals/v1/deal/paged",
+    "deals_properties":     "/crm/v3/properties/deals",
+    "deals_all":            "/crm/v3/objects/deals",
     "deals_recent":         "/deals/v1/deal/recent/modified",
     "deals_detail":         "/deals/v1/deal/{deal_id}",
 
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -72,8 +72,8 @@ ENDPOINTS = {
     "companies_detail":     "/companies/v2/companies/{company_id}",
     "contacts_by_company":  "/companies/v2/companies/{company_id}/vids",
 
-    "deals_properties":     "/crm/v3/properties/deals",
-    "deals_all":            "/crm/v3/objects/deals",
+    "deals_properties":     "/properties/v1/deals/properties",
+    "deals_all":            "/deals/v1/deal/paged",
     "deals_recent":         "/deals/v1/deal/recent/modified",
     "deals_detail":         "/deals/v1/deal/{deal_id}",
 
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -76,7 +76,8 @@ ENDPOINTS = {
     "deals_all":            "/deals/v1/deal/paged",
     "deals_recent":         "/deals/v1/deal/recent/modified",
     "deals_detail":         "/deals/v1/deal/{deal_id}",
-
+    "deals_v3_search":      "/crm/v3/objects/deals/search",
+    "deals_v3_properties":  "/crm/v3/properties/deals",
     "deal_pipelines":       "/deals/v1/pipelines",
 
     "campaigns_all":        "/email/public/v1/campaigns/by-id",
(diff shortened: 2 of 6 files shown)


tests/test_hubspot_bookmarks1.py
```diff
@@ -156,7 +156,7 @@ def expected_bookmarks(self):
 
 
     def get_properties(self):
-        return {'start_date' : '2017-05-01 00:00:00'}
+        return {'start_date' : '2017-05-01T00:00:00Z'}
```
luandy64, line 159: This is because we updated the Circle image to `tap-tester-v4`

tests/test_hubspot_bookmarks2.py
```diff
@@ -104,7 +104,7 @@ def expected_bookmarks(self):
 
 
     def get_properties(self):
-        return {'start_date' : '2017-05-01 00:00:00'}
+        return {'start_date' : '2017-05-01T00:00:00Z'}
```
luandy64, line 107: This is because we updated the Circle image to `tap-tester-v4`

tap_hubspot/tests/test_deals.py
```diff
@@ -23,7 +23,7 @@ def test_associations_are_not_validated(self,
 
         expected_param = {'count': 250, 'includeAssociations': False, 'properties': []}
 
-        mocked_gen_request.assert_called_once_with(ANY, ANY, ANY, expected_param, ANY, ANY, ANY, ANY)
+        mocked_gen_request.assert_called_once_with(ANY, ANY, ANY, expected_param, ANY, ANY, ANY, ANY, v3_fields=ANY)
```
luandy64, line 26: We updated the signature, so this change needed to happen

setup.py
```diff
@@ -15,8 +15,13 @@
           'requests==2.20.0',
           'backoff==1.3.2',
           'requests_mock==1.3.0',
-          'nose'
       ],
+      extras_require= {
+          'dev': [
+              'pylint==2.5.3',
```
luandy64, line 21: I pinned this because I saw failures in the standard list of disables on `pylint v2.6.0`

setup.py
```diff
@@ -15,8 +15,13 @@
...
+              'nose==1.3.7',
```
luandy64, line 22: Added this here so that we don't install it in the Circle file

tap_hubspot/__init__.py
```diff
@@ -307,8 +320,37 @@ def lift_properties_and_versions(record):
             record['properties_versions'] += versions
     return record
 
+def post_search_endpoint(url, data, params=None):
+    params = params or {}
+    hapikey = CONFIG['hapikey']
+    if hapikey is None:
+        if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
+            acquire_access_token_from_refresh_token()
```
asaf-erlich, line 328: So this method fills the access token into the global CONFIG variable? Any reason not to return the access token? Looks weird to me that you call a method here and it's not entirely clear what it's doing. This isn't a big deal...

tap_hubspot/__init__.py
```diff
@@ -307,8 +320,37 @@ def lift_properties_and_versions(record):
...
+    else:
+        params['hapikey'] = hapikey
+        headers = {}
+
+    if 'user_agent' in CONFIG:
+        headers['User-Agent'] = CONFIG['user_agent']
+
+    headers['content-type'] = "application/json"
+
+    # TODO: fix log message
```
asaf-erlich, line 339: What is this TODO about?

tap_hubspot/__init__.py
```diff
@@ -319,6 +361,19 @@ def gen_request(STATE, tap_stream_id, url, params, path, more_key, offset_keys,
...
 
+            if v3_fields:
+                url = get_url('deals_v3_search')
+                body = {"properties": v3_fields, "limit": 100}
+                v3_data = post_search_endpoint(url, body)
+                additional_fields = {}
+                for item in v3_data.json()['results']:
+                    # We nest `y` under 'value' in order to match the
+                    # schema and the shape of other fields in 'properties'
+                    additional_fields[int(item['id'])] = {x:{'value': y} for x,y in item['properties'].items()
```
asaf-erlich, line 372: Can you please use anything other than x, y? Like even key, value? It's hard to follow this comprehension in the way it's written right now.

tap_hubspot/__init__.py
```diff
@@ -319,6 +361,19 @@ def gen_request(STATE, tap_stream_id, url, params, path, more_key, offset_keys,
...
+                                                          if ('hs_date_entered' in x or 'hs_date_exited' in x)}
```
asaf-erlich, line 373: I'm slightly confused. For every key in item['properties'] check if 'hs_date_entered' is "in" the key? Is the key a list of strings? Or is it a long string and 'hs_date_entered' could be somewhere inside that string?

luandy64: > 'hs_date_entered' could be somewhere inside that string

This one

tap_hubspot/__init__.py
luandy64: This is how the other request function did it, so we just copied it in hopes that it wouldn't be too confusing.

> So this method fills the access token into the global CONFIG variable?

Yes, exactly. And the global `CONFIG` gets checked before the tap makes any request

KAllan357 (approved):

tap_hubspot/__init__.py
asaf-erlich: ok

asaf-erlich (approved):

PR #132, opened 2020-11-19
Add `hs_date_entered_*` and `hs_date_exited_*` fields for Deals

luandy64:
# Description of change
This PR hits the V3 Deals CRM API to get all `hs_date_entered_*` and `hs_date_exited_*` fields.

The "batch read" endpoint that Hubspot has lets us search for specific IDs. Given that V1 Deals and V3 Deals have the same primary key (`"dealId"` and `"id"` respectively) values, we can merge these new fields into the existing records

# Manual QA steps
- Ran the tap before and after the change to get records
- Diffed the two sets of records
 
# Risks
- Lower than #124. The main issue there is we did not paginate the V3 response and could not prove whether the first page of V1 records contains the same IDs as the first page of V3 records. But now that we are searching the V3 for explicit IDs we don't have to worry about that
 
# Rollback steps
 - revert this branch and bump the version

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -365,7 +365,7 @@ def process_v3_deals_records(v3_data):
     return transformed_v3_data
 
 def get_v3_deals(v3_fields, v1_data):
-    v1_ids = [{'id': str(record['dealId'])} for record in data[path]]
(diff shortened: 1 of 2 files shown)


luandy64:
Closing this in favor of #133

PR #133, opened 2020-11-20
Add `hs_date_entered_*` and `hs_date_exited_*` fields for the Deals Stream

luandy64:
# Description of change
This PR hits the V3 Deals CRM API to get all `hs_date_entered_*` and `hs_date_exited_*` fields.

The "batch read" endpoint that Hubspot has lets us search for specific IDs. Given that V1 Deals and V3 Deals have the same primary key (`"dealId"` and `"id"` respectively) values, we can merge these new fields into the existing records

# Manual QA steps
- Ran the tap before and after the change to get records
- Diffed the two sets of records
 
# Risks
- Lower than #124. The main issue there is we did not paginate the V3 response and could not prove whether the first page of V1 records contains the same IDs as the first page of V3 records. But now that we are searching the V3 for explicit IDs we don't have to worry about that
 
# Rollback steps
 - revert this branch and bump the version

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -165,7 +165,6 @@ def get_field_schema(field_type, extras=False):
         }
 
 def parse_custom_schema(entity_name, data):
-
     return {
         field['name']: get_field_schema(field['type'], entity_name != 'contacts')
         for field in data
@@ -266,17 +265,13 @@ def parse_source_from_url(url):
         return match.group(1)
     return None
 
-
-@backoff.on_exception(backoff.constant,
-                      (requests.exceptions.RequestException,
-                       requests.exceptions.HTTPError),
-                      max_tries=5,
-                      jitter=None,
-                      giveup=giveup,
-                      on_giveup=on_giveup,
-                      interval=10)
-def request(url, params=None):
-
+def get_params_and_headers(params):
+    """
+    This function makes a params object and headers object based on the
+    authentication values available. If there is an `hapikey` in the config, we
+    need that in `params` and not in the `headers`. Otherwise, we need to get an
+    `access_token` to put in the `headers` and not in the `params`
+    """
     params = params or {}
     hapikey = CONFIG['hapikey']
     if hapikey is None:
@@ -289,6 +284,21 @@ def request(url, params=None):
 
     if 'user_agent' in CONFIG:
         headers['User-Agent'] = CONFIG['user_agent']
+
+    return params, headers
+
+
+@backoff.on_exception(backoff.constant,
+                      (requests.exceptions.RequestException,
+                       requests.exceptions.HTTPError),
+                      max_tries=5,
+                      jitter=None,
+                      giveup=giveup,
+                      on_giveup=on_giveup,
+                      interval=10)
+def request(url, params=None):
+
+    params, headers = get_params_and_headers(params)
 
     req = requests.Request('GET', url, params=params, headers=headers).prepare()
     LOGGER.info("GET %s", req.url)
@@ -322,19 +332,8 @@ def lift_properties_and_versions(record):
     return record
 
 def post_search_endpoint(url, data, params=None):
-    params = params or {}
-    hapikey = CONFIG['hapikey']
-    if hapikey is None:
-        if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
-            acquire_access_token_from_refresh_token()
-        headers = {'Authorization': 'Bearer {}'.format(CONFIG['access_token'])}
-    else:
-        params['hapikey'] = hapikey
-        headers = {}
 
-    if 'user_agent' in CONFIG:
-        headers['User-Agent'] = CONFIG['user_agent']
-
+    params, headers = get_params_and_headers(params)
     headers['content-type'] = "application/json"
 
     with metrics.http_request_timer(url) as timer:
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -76,7 +76,7 @@ ENDPOINTS = {
     "deals_all":            "/deals/v1/deal/paged",
     "deals_recent":         "/deals/v1/deal/recent/modified",
     "deals_detail":         "/deals/v1/deal/{deal_id}",
-    "deals_v3_search":      "/crm/v3/objects/deals/search",
+    "deals_v3_batch_read":  "/crm/v3/objects/deals/batch/read",
     "deals_v3_properties":  "/crm/v3/properties/deals",
     "deal_pipelines":       "/deals/v1/pipelines",
 
@@ -348,6 +348,39 @@ def post_search_endpoint(url, data, params=None):
 
     return resp
 
+def merge_responses(v1_data, v3_data):
+    for v1_record in v1_data:
+        v1_id = v1_record.get('dealId')
+        for v3_record in v3_data:
+            v3_id = v3_record.get('id')
+            if str(v1_id) == v3_id:
+                v1_record['properties'] = {**v1_record['properties'],
+                                           **v3_record['properties']}
+
+def process_v3_deals_records(v3_data):
+    """
+    This function:
+    1. filters out fields that don't contain 'hs_date_entered_*' and
+       'hs_date_exited_*'
+    2. changes a key value pair in `properties` to a key paired to an
+       object with a key 'value' and the original value
+    """
+    transformed_v3_data = []
+    for record in v3_data:
+        new_properties = {field_name : {'value': field_value}
+                          for field_name, field_value in record['properties'].items()
+                          if 'hs_date_entered' in field_name or 'hs_date_exited' in field_name}
+        transformed_v3_data.append({**record, 'properties' : new_properties})
+    return transformed_v3_data
+
+def get_v3_deals(v3_fields, v1_data):
+    v1_ids = [{'id': str(record['dealId'])} for record in data[path]]
+    v3_body = {'inputs': v1_ids,
+               'properties': v3_fields,}
+    v3_url = get_url('deals_v3_batch_read')
+    v3_resp = post_search_endpoint(v3_url, v3_body)
+    return v3_resp.json()['results']
+
 #pylint: disable=line-too-long
 def gen_request(STATE, tap_stream_id, url, params, path, more_key, offset_keys, offset_targets, v3_fields=None):
     if len(offset_keys) != len(offset_targets):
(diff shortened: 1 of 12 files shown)


tests/base.py
```diff
@@ -195,6 +195,10 @@ def create_connection(self, original_properties: bool = True):
         menagerie.verify_check_exit_status(self, exit_status, check_job_name)
         return conn_id
 
+    def ensure_connection(self, original=True):
+        conn_id = connections.ensure_connection(self, original_properties = original)
```
asaf-erlich, line 199: What is the benefit of this method? Is it simply to make original a default true value? Why not fix that in the connections.py file inside of tap tester?

tap_hubspot/__init__.py
```diff
@@ -159,17 +162,23 @@ def get_field_schema(field_type, extras=False):
             }
         }
 
-def parse_custom_schema(entity_name, data):
+def parse_custom_schema(entity_name, data, force_extras=None):
+    if force_extras is not None:
+        extras = force_extras
+    else:
+        extras = entity_name != 'contacts'
     return {
```
asaf-erlich, line 170: Nitpick, but can you add a line break before the return here?

tap_hubspot/__init__.py
```diff
@@ -159,17 +162,23 @@ def get_field_schema(field_type, extras=False):
             }
         }
 
-def parse_custom_schema(entity_name, data):
+def parse_custom_schema(entity_name, data, force_extras=None):
+    if force_extras is not None:
+        extras = force_extras
+    else:
+        extras = entity_name != 'contacts'
```
asaf-erlich, line 169: This feels weird the way the code is written. Why not make force_extras a boolean instead of an Optional of... I guess a boolean? Have I mentioned I miss type hints in python? I wish we used them... but that's not important right now. I see... you allow force extras to be passed as either true or false. 

Is `entity_name != 'contacts'` really the most common default use case? I would rather you just didn't have a default value and passed it in, you only us this method twice, ever right? So what's the benefit of pushing the logic into the method and not from the caller?

tap_hubspot/__init__.py
```diff
@@ -191,6 +200,12 @@ def load_schema(entity_name):
             "properties": custom_schema,
         }
 
+        if entity_name in ["deals"]:
+            v3_schema = get_v3_schema(entity_name)
+            for key, value in v3_schema.items():
+                if 'hs_date_entered' in key or 'hs_date_exited' in key or 'hs_time_in' in key:
```
asaf-erlich, line 206: I'm not 100% sure, but could you use `any` to make this line a bit easier on the eyes? It's essentially saying `are any of the strings in key` right?

tap_hubspot/__init__.py
luandy64: > Is entity_name != 'contacts' really the most common default use case?

Yes, it's what it was before. I didn't spend too much thing clarifying the logic, but I needed to the ability to override that value

tap_hubspot/__init__.py
luandy64: No, it's checking for any of the three prefixes in `key`

tap_hubspot/__init__.py
```diff
@@ -307,8 +322,70 @@ def lift_properties_and_versions(record):
             record['properties_versions'] += versions
     return record
 
+def post_search_endpoint(url, data, params=None):
+    params = params or {}
+    hapikey = CONFIG['hapikey']
+    if hapikey is None:
+        if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():
+            acquire_access_token_from_refresh_token()
```
asaf-erlich, line 330: I slightly remember discussing this in a previous PR... but why not make the code simply:
`headers = {'Authorization': 'Bearer {}'.format(access_token())}`

And then pushing this logic of `if CONFIG['token_expires'] is None or CONFIG['token_expires'] < datetime.datetime.utcnow():` into an access_token method which would take care of it? It feels like a leaky abstraction right now that seems unnecessary.

tap_hubspot/__init__.py
```diff
@@ -307,8 +322,70 @@ def lift_properties_and_versions(record):
...
+    1. filters out fields that don't contain 'hs_date_entered_*' and
+       'hs_date_exited_*'
+    2. changes a key value pair in `properties` to a key paired to an
+       object with a key 'value' and the original value
+    """
+    transformed_v3_data = []
+    for record in v3_data:
+        new_properties = {field_name : {'value': field_value}
+                          for field_name, field_value in record['properties'].items()
+                          if 'hs_date_entered' in field_name or 'hs_date_exited' in field_name or 'hs_time_in' in field_name}
```
asaf-erlich, line 374: Again I think `any` could be used to reduce the number of `if` and `or` in this line...

tap_hubspot/__init__.py
```diff
@@ -319,6 +396,14 @@ def gen_request(STATE, tap_stream_id, url, params, path, more_key, offset_keys,
         while True:
             data = request(url, params).json()
 
+            if v3_fields:
+                v3_data = get_v3_deals(v3_fields, data[path])
+
+                # The shape of v3_data is different than the V1 response,
+                # so we transform v3 to look like v1
```
asaf-erlich, line 403: The code is very clean, but I kind of hate that we have to do this lol

tap_hubspot/__init__.py
```diff
@@ -522,9 +609,14 @@ def sync_deals(STATE, ctx):
         params['includeAllProperties'] = True
         params['allPropertiesFetchMode'] = 'latest_version'
 
+        # Grab selected `hs_date_entered/exited` fields to call the v3 endpoint with
+        v3_fields = [x[1].replace('property_', '')
+                     for x,y in mdata.items() if x and (y.get('selected') == True or has_selected_properties)
```
asaf-erlich, line 614: Niptick, but please do not name these x,y. Name them something longer so I understand better what information they represent.

tests/test_hubspot_automatic_fields_test.py
```diff
@@ -0,0 +1,132 @@
...
+        exit_status = menagerie.get_exit_status(conn_id, check_job_name)
+        menagerie.verify_check_exit_status(self, exit_status, check_job_name)
+
+        found_catalogs = menagerie.get_catalogs(conn_id)
+        self.assertGreater(len(found_catalogs), 0, msg="unable to locate schemas for connection {}".format(conn_id))
+
+        # Select only the expected streams tables
+        expected_streams = self.expected_streams()
+        catalog_entries = [ce for ce in found_catalogs if ce['tap_stream_id'] in expected_streams]
+        self.select_all_streams_and_fields(conn_id, catalog_entries, select_all_fields=False)
```
asaf-erlich, line 37: With the exception of this line, the majority of this code above and some of the test code below is identical. Can you move them into the base? You can look at tap-square for examples. In my opinion most of this should be moved to be shared tap-tester code... since all our tests do it.

* run_and_verify_check_mode: https://github.com/singer-io/tap-square/blob/master/tests/base.py#L513
* run_and_verify_sync: https://github.com/singer-io/tap-square/blob/master/tests/base.py#L573

tap_hubspot/__init__.py
asaf-erlich: Ok, but why do you need to put this logic inside the method. I see this method called twice below. Why not just pass in `extras = entity_name != 'contacts'` as the second required parameter to this method for the only time it's not set to `extras=False` in the method `get_v3_schema` below?

tap_hubspot/__init__.py
asaf-erlich: if it's a prefix why not do `startswith` (I think that's the method in python)? And why does that mean you can't use `any`?

https://stackoverflow.com/questions/3389574/check-if-multiple-strings-exist-in-another-string

tap_hubspot/__init__.py
luandy64: I noticed that the one time I use `force_extras`, the boolean would come out to `false` anyway. So I took out this addition

tap_hubspot/__init__.py
luandy64: I'm leaving this since it's the first line now

tap_hubspot/__init__.py
luandy64: I cleaned this up with an `any()`

tap_hubspot/__init__.py
luandy64: I moved this into a function we use in both functions now

tap_hubspot/__init__.py
luandy64: Done

tap_hubspot/__init__.py
luandy64: Done

tap_hubspot/__init__.py
```diff
@@ -319,6 +394,14 @@ def gen_request(STATE, tap_stream_id, url, params, path, more_key, offset_keys,
         while True:
             data = request(url, params).json()
 
+            if v3_fields:
```
KAllan357, line 397: In a few spots below here, we call `data[path]`. Should we check that this exists and if not, raise an exception? That would make this code slightly safer and provide a more obvious failure.

KAllan357 (approved):

tap_hubspot/__init__.py
asaf-erlich: Sure

tap_hubspot/__init__.py
```diff
@@ -191,6 +198,12 @@ def load_schema(entity_name):
             "properties": custom_schema,
         }
 
+        if entity_name in ["deals"]:
+            v3_schema = get_v3_schema(entity_name)
+            for key, value in v3_schema.items():
+                if any(prefix in key for prefix in V3_PREFIXES):
```
asaf-erlich, line 204: Nice

tap_hubspot/__init__.py
```diff
@@ -252,17 +265,13 @@ def parse_source_from_url(url):
...
-def request(url, params=None):
-
+def get_params_and_headers(params):
+    """
+    This function makes a params object and headers object based on the
+    authentication values available. If there is an `hapikey` in the config, we
+    need that in `params` and not in the `headers`. Otherwise, we need to get an
+    `access_token` to put in the `headers` and not in the `params`
+    """
     params = params or {}
```
asaf-erlich, line 275: feels like this could also just be a default value, but unfortunately dict default values can be modified which linters catch... so just leave it like this...

tap_hubspot/__init__.py
```diff
@@ -276,6 +285,21 @@ def request(url, params=None):
...
+                      (requests.exceptions.RequestException,
+                       requests.exceptions.HTTPError),
+                      max_tries=5,
+                      jitter=None,
+                      giveup=giveup,
+                      on_giveup=on_giveup,
+                      interval=10)
+def request(url, params=None):
+
+    params, headers = get_params_and_headers(params)
```
asaf-erlich, line 301: I like how clean this looks to me now

tap_hubspot/__init__.py
```diff
@@ -307,8 +331,59 @@ def lift_properties_and_versions(record):
             record['properties_versions'] += versions
     return record
 
+def post_search_endpoint(url, data, params=None):
```
asaf-erlich, line 334: same with this entire method

tap_hubspot/__init__.py
```diff
@@ -307,8 +331,59 @@ def lift_properties_and_versions(record):
...
+def get_v3_deals(v3_fields, v1_data):
+    v1_ids = [{'id': str(record['dealId'])} for record in v1_data]
+
+    v3_body = {'inputs': v1_ids,
+               'properties': v3_fields,}
+    v3_url = get_url('deals_v3_batch_read')
+    v3_resp = post_search_endpoint(v3_url, v3_body)
+    return v3_resp.json()['results']
+
 #pylint: disable=line-too-long
```
asaf-erlich, line 385: Not important for this PR but I normally add line-too-long to the disables in the pylint command. It's from a time when screens were not so large...

tap_hubspot/__init__.py
```diff
@@ -499,7 +589,7 @@ def sync_deals(STATE, ctx):
     max_bk_value = start
     LOGGER.info("sync_deals from %s", start)
     most_recent_modified_time = start
-    params = {'count': 250,
+    params = {'limit': 100,
```
asaf-erlich, line 592: I didn't ask last time, but is there a reason we dropped this to paginate by 100 and not by 250?

asaf-erlich (approved):

tap_hubspot/__init__.py
luandy64: The V3 endpoint has a limit of 100. So I needed the V1 endpoint to match it in order for my "search for this deal IDs" strategy to work without pagination

PR #150, opened 2021-03-21
Allow skipping of property extras and versions

pankajsaini-ldt:
# Description of change
This change adds an optional config in config.json named `skip_property_extras_and_versions`, which allows tap users to select only the recent value of the properties from hubspot objects and skip any extras like timestamp, source, sourceId as well as the versions. The extras and version fields create an issue because of large number of columns created when using this tap with [target-bigquery](https://github.com/adswerve/target-bigquery). This option also allows the users to select only the recent data from hubspot if they don't require the versions and additional metadata related with the field values.

# Manual QA steps
- Run tap-hubspot in discovery mode and generate catalog
- Run tap-hubspot in sync mode. You will notice that objects like deals and companies that have `properties_` attributes will give the actual values of the properties instead of returning an object with value and other metadata (e.g. timestamp, source, sourceId)
 
# Risks
 - This change requires the discovery of the catalog again. So if you are running it in an existing pipeline, ensure that you update your catalog (properties) file.
 
# Rollback steps
 - revert this branch

### DIFF
--- a/config.sample.json
+++ b/config.sample.json
@@ -4,5 +4,6 @@
   "client_secret": "my_secret",
   "refresh_token": "my_token",
   "start_date": "2017-01-01T00:00:00Z",
-  "disable_collection": false
+  "disable_collection": false,
+  "skip_property_extras_and_versions": false
 }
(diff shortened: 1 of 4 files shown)


pankajsaini-ldt:
Closed PR in order to avoid merging additional changes that were made on the incoming branch. I will add a new PR after I add changes on a dedicated branch.

PR #158, opened 2021-08-18
Qa/automatic fields increase coverage

yusaku2:
# Description of change
https://jira.talendforge.org/browse/TDL-14763
Added missing streams to test and created workarounds in order to handle streams with missing key issue.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_hubspot_automatic_fields_test.py
+++ b/tests/test_hubspot_automatic_fields_test.py
@@ -63,6 +63,7 @@ class TestHubspotAutomaticFields(HubspotBaseTest):
             with self.subTest(stream=stream):
                 data = synced_records.get(stream)
 
+                # TODO Remove this, it could cover up streams that are not actually tested 
                 if not data:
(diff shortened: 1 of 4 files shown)


kspeer825 (approved):

tests/test_hubspot_automatic_fields_test.py
```diff
@@ -11,19 +11,12 @@ def name(self):
...
-            'subscription_changes',
-            'email_events',
-            'companies',
-            'contacts',
-            'deals',
-            'engagements',
-            'contacts_by_company',
-        })
+        """streams to test"""
+        return self.expected_check_streams()
```
kspeer825, line 19: If all streams are under test we should get rid of this method and call expected_check_streams directly in the test.

kspeer825 (approved):
Looks great, just 1 minor change

PR #159, opened 2021-08-23
add a test to check for child streams without parent streams selected

HarrisonMarcRose:
# Description of change
Implement https://jira.talendforge.org/browse/TDL-14966

When a child stream is selected without its parent (usually not allowed via field selection but can be in this case) it will cause a critical error.  We must ensure the error has a message to the user that is sufficient that they can resolve the problem on their own be selecting the parent stream

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_hubspot_child_stream_only.py
+++ b/tests/test_hubspot_child_stream_only.py
@@ -1,0 +1,72 @@
+"""Test tap discovery mode and metadata/annotated-schema."""
+import re
+
+from tap_tester import connections
+from tap_tester import menagerie
+from tap_tester import runner
+
+from base import HubspotBaseTest
+
+
+class FieldSelectionChildTest(HubspotBaseTest):
+    """Test tap field selection of child streams without its parent."""
+
+    @staticmethod
+    def name():
+        return "tap_tester_hubspot_child_streams_test"
+
+    def test_run(self):
+        """
+        Verify that when a child stream is selected without its parent that
+        • a critical error in the tap occurs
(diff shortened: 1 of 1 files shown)


tests/test_hubspot_child_stream_only.py
```diff
@@ -0,0 +1,72 @@
+"""Test tap discovery mode and metadata/annotated-schema."""
```
kspeer825, line 1: This is not applicable

tests/test_hubspot_child_stream_only.py
```diff
@@ -0,0 +1,72 @@
...
+        sync_job_name = runner.run_sync_mode(self, conn_id)
+
+        # Verify tap and target exit codes
+        exit_status = menagerie.get_exit_status(conn_id, sync_job_name)
+        
+        # Verify that the tap error message shows you need to select the parent stream
+        self.assertEqual(exit_status['tap_error_message'],
+                         ('Unable to extract contacts_by_company data. '
+                          'To receive contacts_by_company data, you also need to select companies.'))
+        self.assertRaises(AssertionError, menagerie.verify_sync_exit_status, self, exit_status, sync_job_name)
```
kspeer825, line 52: nice

kspeer825: Maybe these two assertions should swap order.

kspeer825 (approved):
Just some minor comments

PR #162, opened 2021-08-27
Qa/test client posts

yusaku2:
# Description of change
https://jira.talendforge.org/browse/TDL-14774
Able to create records for HubSpot streams through GETs and POSTs.
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -61,7 +61,7 @@ class TestClient():
         json_response = response.json()
 
         return json_response
-    
+
     @backoff.on_exception(backoff.constant,
                           (requests.exceptions.RequestException,
                            requests.exceptions.HTTPError),
@@ -481,7 +481,7 @@ class TestClient():
                  }
                ]
              }
-        
+
         # generate a contacts record
         response = self.post(url, data)
         records = [response]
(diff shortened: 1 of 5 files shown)


tests/client.py
```diff
@@ -403,6 +425,501 @@ def get_workflows(self):
...
+                        "fromObjectId": company_id,
+                        "toObjectId": contact_id,
+                        "category": "HUBSPOT_DEFINED",
+                        "definitionId": 2
+                    
+                    }                    
+                    # generate a record
+                    self.put(url, data)
+                    records = [{'company-id': company_id, 'contact-id': contact_id}]
+                    return records
```
luandy64, line 577: Can we get in a state where there are two `contact_id`s that are not in our list of `contact_by_company` ids?

tests/client.py
```diff
@@ -403,6 +425,501 @@ def get_workflows(self):
...
+        """
+        record_uuid = str(uuid.uuid4()).replace('-', '')
+
+        url = f"{BASE_URL}"
+        data = {}
+        
+        # generate a record
+        response = self.post(url, data)
+        records = [response]
+        return records
```
luandy64, line 502: Does this `TODO` mean it's not implemented?

If so, can we be consistent and `raise NotImplementedError()`?
If not, then can we remove the `TODO`?

tests/client.py
```diff
@@ -403,6 +425,501 @@ def get_workflows(self):
...
+                    "subscribed": True,
+                    "optState": "OPT_IN",
+                    "legalBasis": "PERFORMANCE_OF_CONTRACT",
+                    "legalBasisExplanation": "We need to send them these emails as part of our agreement with them."
+                }
+            ]
+        }
+        
+        # generate a record
+        response = self.put(url.format(record_uuid+"@stitchdata.com"), data)
```
luandy64, line 874: Feel free to ignore this, it's nit-picky.

> url.format(record_uuid+"@stitchdata.com"

I don't like this `format()` here. We already know `record_uuid` by the time we initialize `url`.

So I think this should be done on line 860 instead

luandy64 (changes requested):
I'm going to block this until we talk about the `create_contacts_by_company()` comment that I left. Otherwise this looks good

luandy64 (approved):

PR #165, opened 2021-09-07
updated expected records for companies to filter after getall

yusaku2:
# Description of change
Companies stream was not getting the correct expected records because we were using recent companies API instead of getting all companies, then filtering.
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_hubspot_all_fields_test.py
+++ b/tests/test_hubspot_all_fields_test.py
@@ -1,7 +1,7 @@
 import tap_tester.connections as connections
 import tap_tester.menagerie   as menagerie
 import tap_tester.runner      as runner
-
+import datetime
 from base import HubspotBaseTest
 from client import TestClient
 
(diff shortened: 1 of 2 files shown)


tests/test_hubspot_all_fields_test.py
```diff
@@ -96,11 +98,20 @@ def testable_streams(self):
...
+        return {'start_date' : '2021-08-05T00:00:00Z'}
+
+    # TODO move the overriden start date up as much as possible to minimize the run time
+    #      it can probably be dynamic like today minus 7 days
 
     @classmethod
     def setUpClass(cls):
         cls.maxDiff = None  # see all output in failure
 
+        # TODO my_timestamp needs to be driven off of start_date
```
kspeer825, line 111: Oh we should do this so the run time does not grow each day.

tests/test_hubspot_all_fields_test.py
```diff
@@ -131,12 +142,12 @@ def test_run(self):
 
         found_catalogs = self.run_and_verify_check_mode(conn_id)
 
-        # moving the state up so the sync will be shorter and the test takes less time
-        state = {'bookmarks': {'companies': {'current_sync_start': None,
-                                             'hs_lastmodifieddate': self.my_timestamp,
-                                             'offset': {}}},
-                 'currently_syncing': None}
-        menagerie.set_state(conn_id, state)
+        # # moving the state up so the sync will be shorter and the test takes less time
```
kspeer825, line 145: This can be removed altogether

kspeer825 (changes requested):

tests/test_hubspot_all_fields_test.py
yusaku2: will do this when test and data is more stable

kspeer825 (approved):

PR #195, opened 2022-06-07
TDL-19319 Make replication key as automatic inclusion

prijendev:
# Description of change
- Replaced replication key of deals stream with `property_hs_lastmodifieddate` field.
- Updated replication method of `deals` stream in the tap to `INCREMENTAL`.
- Updated replication key in the `base.py` of tap-tester.
- Updated `automatic_field` and `start_date` tap tester test case.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -599,6 +599,7 @@ def sync_companies(STATE, ctx):
 def has_selected_custom_field(mdata):
     top_level_custom_props = [x for x in mdata if len(x) == 2 and 'property_' in x[1]]
     for prop in top_level_custom_props:
+        # Return 'True' if the custom field is automatic.
         if (mdata.get(prop, {}).get('selected') == True) or (mdata.get(prop, {}).get('inclusion') == "automatic"):
             return True
     return False
@@ -612,7 +613,7 @@ def sync_deals(STATE, ctx):
     # `hs_lastmodifieddate` is available in the properties field at the nested level.
     # As `hs_lastmodifieddate` is not available at the 1st level it can not be marked as automatic inclusion.
     # tap includes all nested fields of the properties field as custom fields in the schema by appending the
-    # prefix property_ along with each field.
+    # prefix `property_` along with each field.
     # That's why bookmark_key is `property_hs_lastmodifieddate` so that we can mark it as automatic inclusion.
 
     bookmark_field_in_record = 'hs_lastmodifieddate'
@@ -964,6 +965,7 @@ STREAMS = [
     Stream('subscription_changes', sync_subscription_changes, ['timestamp', 'portalId', 'recipient'], 'startTimestamp', 'INCREMENTAL'),
     Stream('email_events', sync_email_events, ['id'], 'startTimestamp', 'INCREMENTAL'),
     Stream('contacts', sync_contacts, ["vid"], 'versionTimestamp', 'INCREMENTAL'),
+    Stream('deals', sync_deals, ["dealId"], 'property_hs_lastmodifieddate', 'INCREMENTAL'),
 
     # Do these last as they are full table
     Stream('forms', sync_forms, ['guid'], 'updatedAt', 'FULL_TABLE'),
@@ -972,7 +974,6 @@ STREAMS = [
     Stream('campaigns', sync_campaigns, ["id"], None, 'FULL_TABLE'),
     Stream('contact_lists', sync_contact_lists, ["listId"], 'updatedAt', 'FULL_TABLE'),
     Stream('companies', sync_companies, ["companyId"], 'hs_lastmodifieddate', 'FULL_TABLE'),
-    Stream('deals', sync_deals, ["dealId"], 'property_hs_lastmodifieddate', 'INCREMENTAL'),
     Stream('deal_pipelines', sync_deal_pipelines, ['pipelineId'], None, 'FULL_TABLE'),
     Stream('engagements', sync_engagements, ["engagement_id"], 'lastUpdated', 'FULL_TABLE')
 ]
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -6,13 +6,13 @@ import os
 import re
 import sys
 import json
-
+# pylint: disable=import-error
 import attr
 import backoff
 import requests
 import singer
 import singer.messages
-import singer.metrics as metrics
+from singer import metrics
 from singer import metadata
 from singer import utils
 from singer import (transform,
@@ -134,7 +134,7 @@ def clean_state(state):
     """ Clear deprecated keys out of state. """
     for stream, bookmark_map in state.get("bookmarks", {}).items():
         if "last_sync_duration" in bookmark_map:
-            LOGGER.info("{} - Removing last_sync_duration from state.".format(stream))
+            LOGGER.info("%s - Removing last_sync_duration from state.", stream)
             state["bookmarks"][stream].pop("last_sync_duration", None)
 
 def get_url(endpoint, **kwargs):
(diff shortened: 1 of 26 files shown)


tap_hubspot/__init__.py
```diff
@@ -599,14 +599,23 @@ def sync_companies(STATE, ctx):
 def has_selected_custom_field(mdata):
     top_level_custom_props = [x for x in mdata if len(x) == 2 and 'property_' in x[1]]
     for prop in top_level_custom_props:
-        if mdata.get(prop, {}).get('selected') == True:
+        if (mdata.get(prop, {}).get('selected') == True) or (mdata.get(prop, {}).get('inclusion') == "automatic"):
```
prijendev, line 602: Return `True` if the custom field is automatic.

dbshah1212 (approved):

tap_hubspot/__init__.py
```diff
@@ -963,7 +972,7 @@ class Stream(object):
     Stream('campaigns', sync_campaigns, ["id"], None, 'FULL_TABLE'),
     Stream('contact_lists', sync_contact_lists, ["listId"], 'updatedAt', 'FULL_TABLE'),
     Stream('companies', sync_companies, ["companyId"], 'hs_lastmodifieddate', 'FULL_TABLE'),
-    Stream('deals', sync_deals, ["dealId"], 'hs_lastmodifieddate', 'FULL_TABLE'),
+    Stream('deals', sync_deals, ["dealId"], 'property_hs_lastmodifieddate', 'INCREMENTAL'),
```
NevilParikh14, line 975: Can you please move this under https://github.com/singer-io/tap-hubspot/pull/195/files#diff-fa09b051e2cdacf5d892f6da3664b221836c73217ae4f4b8ef2b87ccbbca220dR963

namrata270998 (approved):

karanpanchal-crest (changes requested):

tap_hubspot/__init__.py
prijendev: Added comment in the code.

tap_hubspot/__init__.py
prijendev: Updated

NevilParikh14 (approved):

tests/base.py
```diff
@@ -105,7 +105,7 @@ def expected_metadata(self):  # DOCS_BUG https://stitchdata.atlassian.net/browse
             "deals": {
                 self.PRIMARY_KEYS: {"dealId"},
                 self.REPLICATION_METHOD: self.INCREMENTAL,
-                self.REPLICATION_KEYS: {"hs_lastmodifieddate"},
+                self.REPLICATION_KEYS: {"property_hs_lastmodifieddate"},
```
kspeer825, line 108: Should this include the `property_` prefix from the denesting method? None of the other replicaiton keys have this. And I think it will break every customer's connection that is replicating this stream.

prijendev: Currently, `hs_lastmodifieddate` is the bookmark key for the `deals` stream. `hs_lastmodifieddate` is available in the properties field at the nested level. 
For example, the response structure:
```
{
    "dealId": 1,
    "properties": {
        "hs_lastmodifieddate": 1653590080602
    }
}
```

As `hs_lastmodifieddate` is not available at the 1st level, it can not be marked as automatic inclusion and we can not select that particular field even. Also, tap includes all nested fields of the properties field as custom fields in the schema by appending the prefix `property_` along with each field.

For example, the schema of the deals stream looks like as below,
```
{
    "dealId": {"type": ["null", "integer"]},
    "properties": {"type": "object", "properties": {
            "hs_lastmodifieddate": {},
            "hs_likelihood_to_close": {}
        }
    },
    "property_hs_lastmodifieddate": {},
    "property_hs_likelihood_to_close": {}
}
```
That's why we are just marking  `property_hs_lastmodifieddate` as the inclusion of automatic. So, I am not sure in which case it will break the customer's connection because we have not removed or updated any fields.

savan-chovatiya (approved):

tests/base.py
kspeer825: I believe this nested structure is true of the other streams as well, but `companies` for example denests the key and does not prefix `property_` onto the key. I think it will break downstream for customers because the value of the replication key is changing.

prijendev: @kspeer825  We have changed the bookmark key just for the `deals `stream. This change will not affect any other streams. 
If bookmark(`last_modifieddate`) is already available for the deals stream, then tap will use only that value for the first time. (#196 )
After completing sync, the tap will write an updated bookmark in the new bookmark key(i.e. `property_lastmodified_date`).
We haven't removed the existing bookmark as well, so in case we need to revert back, we can able to. 
After sync, the bookmark will look like the below.

```
  "bookmarks": {
        "stream_id_1": {
            "old_bookmark": "2022-04-08T00:00:00",
            "current_bookmark": "2022-04-15T00:00:00"
        }
    }
```

We have incorporated this change to fix the customer's issue. So, @KrisPersonal  shall we plan an alpha release first for the particular customer?

tests/test_hubspot_automatic_fields.py
```diff
@@ -69,7 +69,7 @@ def test_run(self):
                 expected_keys = self.expected_automatic_fields().get(stream)
 
                 # BUG_TDL-9939 https://jira.talendforge.org/browse/TDL-9939 Replication keys are not included as an automatic field for these streams
-                if stream in {'companies', 'deals', 'subscription_changes', 'email_events'}:
+                if stream in {'companies', 'subscription_changes', 'email_events'}:
```
kspeer825, line 72: Can we make this change for the other streams with replication keys that are missing automatic inclusion?

prijendev: @kspeer825 Yes, the replication key is not automatic for companies, subscription_changes, and email_events streams.

1. The `companies` stream has the same issue([link](https://talend.slack.com/archives/C01TGC4QAUE/p1653902238364979?thread_ts=1653593162.728269&cid=C01TGC4QAUE)) of the replication key as the `deals` stream. So, can we create a card to resolve this issue and work on that?
2.  `subscription_changes` and `email_events` streams are retrieving records in the date window. These streams are saving the `start date of the last window` as a bookmark value in `startTimestamp` field.
`email_events` stream has a `created` field in the API response and `subscription_changes` has a `timestamp` field in the response that we may use as a replication key for the respective stream. But, we are not sure whether filter parameters(`Start timestamp`, `End timestamp`) are following these keys or not as we can't find any doc stating the same, so we asked a question for the same in the community([link](https://community.hubspot.com/t5/APIs-Integrations/Clarity-on-parameter-of-API/td-p/657281)).
  **Possible solutions:** 
    1. We can add a new `startTimestamp` field in the record explicitly and mark that field as an automatic to keep current behavior as is.
    2. Update the bookmark key for `email_events` and `subscription_changes` as `created` and `timestamp` respectively if we got a positive reply from the support.

kspeer825: 1. Sounds good 
2. Are you able to verify the usage of these keys by exploratory testing? If you have a record before the StartTimestamp, another record after the EndTimestamp, and another record between those two I think you'd answer that question. I supposed you'd actually need two more records at either of those fence posts to see if the API treats either inclusively.

prijendev: As per discussion with @cosimon, we concluded to make a change in the replication key just for the deals stream. We are not making any changes in `subscription_changes` and `email_events` as of now.

kspeer825 (approved):
Looks good 👍

cosimon (approved):

PR #196, opened 2022-06-09
Utilize bookmark of existing connection

prijendev:
# Description of change

To use bookmark of existing connection we have updated code as below,

- Updated get_start function to use older bookmark key value if it is available for any existing connection.
  - If the current bookmark_key is available in the state, then get_start returns the bookmark_key value.
  - If it is not available then check and return the older_bookmark_key(if it is available) in the state for the existing connection.
  - If non of the key available in the state for a particular stream, then return start_date.
- Added unit test cases for the get_start function.

**Note**: We will not clear out existing bookmarks from the state. Instead, we will keep both bookmark values older and newer. 
# Manual QA steps
 - Run the sync mode without state and verify that the state file contains only property_hs_lastmodifieddate.
 - Run the sync mode with the state file(contains only hs_lastmodifieddate) and verify that the state file contains both hs_lastmodifieddate and property_hs_lastmodifieddate.
    - Also, verify that tap retrieves records based on hs_lastmodifieddate.
 - Run the sync mode with the state file(contains both hs_lastmodifieddate and property_hs_lastmodifieddate) and verify that the state file contains both hs_lastmodifieddate and property_hs_lastmodifieddate. 
   - Also, verify that tap retrieves records based on property_hs_lastmodifieddate.
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -628,9 +628,11 @@ def sync_deals(STATE, ctx):
 
     bookmark_field_in_record = 'hs_lastmodifieddate'
 
-    # Tap was used to write bookmark using replication key `hs_lastmodifieddate`. 
-    # Now, as the replication key gets changed to "property_hs_lastmodifieddate", `get_start` function would return 
-    # bookmark value of older bookmark key(`hs_lastmodifieddate`) if it is available. 
+    # Tap was used to write bookmark using replication key `hs_lastmodifieddate`.
+    # Now, as the replication key gets changed to "property_hs_lastmodifieddate", `get_start` function would return
+    # bookmark value of older bookmark key(`hs_lastmodifieddate`) if it is available.
+    # So, here `older_bookmark_key` is the previous bookmark key that may be available in the state of
+    # the existing connection.
 
     start = utils.strptime_with_tz(get_start(STATE, "deals", bookmark_key, older_bookmark_key=bookmark_field_in_record))
     max_bk_value = start
@@ -687,8 +689,6 @@ def sync_deals(STATE, ctx):
             if not modified_time or modified_time >= start:
                 record = bumble_bee.transform(lift_properties_and_versions(row), schema, mdata)
                 singer.write_record("deals", record, catalog.get('stream_alias'), time_extracted=utils.now())
-
-    STATE = singer.clear_bookmark(STATE, 'deals', 'hs_lastmodifieddate')
 
     STATE = singer.write_bookmark(STATE, 'deals', bookmark_key, utils.strftime(max_bk_value))
     singer.write_state(STATE)
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -673,6 +673,9 @@ def sync_deals(STATE, ctx):
                 record = bumble_bee.transform(lift_properties_and_versions(row), schema, mdata)
                 singer.write_record("deals", record, catalog.get('stream_alias'), time_extracted=utils.now())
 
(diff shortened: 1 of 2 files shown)


dbshah1212 (approved):

tap_hubspot/__init__.py
```diff
@@ -617,7 +627,14 @@ def sync_deals(STATE, ctx):
     # That's why bookmark_key is `property_hs_lastmodifieddate` so that we can mark it as automatic inclusion.
 
     bookmark_field_in_record = 'hs_lastmodifieddate'
```
karanpanchal-crest, line 629: Change the variable name to "last_modified_date" for readability

karanpanchal-crest (changes requested):

tap_hubspot/__init__.py
```diff
@@ -98,9 +98,19 @@ class StateFields:
...
-def get_start(state, tap_stream_id, bookmark_key):
+def get_start(state, tap_stream_id, bookmark_key, older_bookmark_key=None):
+    """
+    If the current bookmark_key is available in the state, then return the bookmark_key value.
+    If it is not available then check and return the older_bookmark_key in the state for the existing connection.
+    If non of the key available in the state for a particular stream, then return start_date.
+    """
     current_bookmark = singer.get_bookmark(state, tap_stream_id, bookmark_key)
     if current_bookmark is None:
+        if older_bookmark_key:
```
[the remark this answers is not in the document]
prijendev: Added comment.

tap_hubspot/tests/unittests/test_get_start.py
```diff
@@ -0,0 +1,94 @@
...
+        # Verify that returned value is start_date
+        self.assertEqual(returned_value, expected_value)
+
+    def test_get_start_with_both_bookmark(self):
+        """
+        This test verifies that the `get_start` function returns current_bookmark from the state
+        if both old and current bookmark is available in the state.
+        """
+
+        mock_state = {
```
karanpanchal-crest, line 80: @prijendev, please make the function "get_state" to attach as many keys in the dictionary

karanpanchal-crest (changes requested):

karanpanchal-crest (approved):

kspeer825 (commented):
Will the old bookmark key be removed from state after syncing and saving the new bookmark key to state?

prijendev:
> Will the old bookmark key be removed from state after syncing and saving the new bookmark key to state?

No, the existing bookmark key will not be removed from the state after sync. Instead, the tap will save the old bookmark key and the new bookmark key in the state.

kspeer825:
> [ View changes ](/singer-io/tap-hubspot/pull/196/files/e416cbf63cc8396f2c4b0f35b5f93bacd4300458)

I don't know if we should be saving an old key and value in state that will never be used. Can this just be replaced instead? Might be good to get some internal dev input here.

prijendev:
> > [ View changes ](/singer-io/tap-hubspot/pull/196/files/e416cbf63cc8396f2c4b0f35b5f93bacd4300458)
> 
> I don't know if we should be saving an old key and value in state that will never be used. Can this just be replaced instead? Might be good to get some internal dev input here.

@kspeer825  I think we should keep the old bookmark key in the state for backward compatibility. We have discussed the same thing with @cosimon and concluded to keep the old bookmark key in the state in case we need to revert back for any client.

kspeer825 (approved):

RushiT0122 (approved):

PR #218, opened 2023-02-03
Tdl 21819 tickets stream dev changes

rdeshmukh15:
# Description of change
(write a short description or paste a link to JIRA)

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -96,6 +96,8 @@ ENDPOINTS = {
     "forms":                "/forms/v2/forms",
     "workflows":            "/automation/v3/workflows",
     "owners":               "/owners/v2/owners",
+
+    "tickets":              "/crm/v4/objects/tickets",
 }
 
 def get_start(state, tap_stream_id, bookmark_key, older_bookmark_key=None):
@@ -676,6 +678,7 @@ def sync_deals(STATE, ctx):
                      and any(prefix in breadcrumb[1] for prefix in V3_PREFIXES)]
 
     url = get_url('deals_all')
+
     with Transformer(UNIX_MILLISECONDS_INTEGER_DATETIME_PARSING) as bumble_bee:
         for row in gen_request(STATE, 'deals', url, params, 'deals', "hasMore", ["offset"], ["offset"], v3_fields=v3_fields):
             row_properties = row['properties']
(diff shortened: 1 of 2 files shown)


tap_hubspot/__init__.py
```diff
@@ -699,7 +702,74 @@ def sync_deals(STATE, ctx):
...
+            # Checking the bookmark value is present and the record is having
+            # grater than or equal to defined previous bookmark value
+            if modified_time and modified_time >= bookmark_value:
+                singer.write_record(stream_id, row, catalog.get(
+                    'stream_alias'), time_extracted=utils.now())
+                max_bk_value = modified_time
+
+    STATE = singer.write_bookmark(STATE, stream_id, bookmark_key, utils.strftime(max_bk_value))
+    singer.write_state(STATE)
+    return STATE
```
kethan1122, line 769: I don't think this return statement is required

kethan1122 (changes requested):

tap_hubspot/__init__.py
```diff
@@ -699,7 +702,74 @@ def sync_deals(STATE, ctx):
...
+    url = get_url(stream_id)
+
+    with Transformer(UNIX_MILLISECONDS_INTEGER_DATETIME_PARSING):
+        for row in gen_request_tickets(STATE, stream_id, url, params, 'results', "paging"):
+
+            modified_time_org = row[bookmark_key]
+            modified_time = utils.strptime_with_tz(datetime.datetime.strptime(
+                modified_time_org, "%Y-%m-%dT%H:%M:%S.%fZ").strftime("%Y-%m-%dT%H:%M:%SZ"))
+
+            # Checking the bookmark value is present and the record is having
```
bryantgray, line 760: Can we reword this to make it a little more clear?

Maybe something like:
> Checking the bookmark value is present on the record and it
is greater than or equal to defined previous bookmark value

tap_hubspot/__init__.py
```diff
@@ -96,6 +96,8 @@ class StateFields:
     "forms":                "/forms/v2/forms",
     "workflows":            "/automation/v3/workflows",
     "owners":               "/owners/v2/owners",
+
+    "tickets":              "/crm/v4/objects/tickets",
```
kethan1122, line 100: Is this `v3` or `v4` ? it is mentioned as `v3` in the provided docs

> To request a list of all tickets, make a GET request to /crm/v3/objects/tickets.

rdeshmukh15: It is `v4`, if you check the API doc, the latest version of the CRM APIS, it will be `v4` 
Here, the requirement is to list the tickets with their associations and for this, I have used the endpoint `/crm/v4/objects/tickets/{ticketId}/associations/{toObjectType} ` mentioned in the endpoint section of [API doc link](https://developers.hubspot.com/docs/api/crm/tickets)

tap_hubspot/__init__.py
rdeshmukh15: `return STATE` is required. If you see line no. 1111, STREAM.sync(STATE, ctx) is expecting the return value `STATE` and if I remove return statement then it will return None and that will cause the issue. Hence, keeping it as it is.

kethan1122 (approved):

kethan1122 (approved):

bryantgray (approved):

RushiT0122 (approved):

PR #220, opened 2023-03-06
Include properties(default + custom) in tickets stream

sgandhi1311:
# Description of change
- In discovery mode, Generate the schema dynamically using the tickets [properties](https://developers.hubspot.com/docs/api/crm/properties) endpoint.
- To get the property details for the tickets stream, the list of property fields needs to be passed in query params. These property fields are shown at the top level for the customers to select them explicitly. 
- In tickets sync mode, the property fields are taken from the catalog, and accordingly the API request URI is formed.

 - (For the existing connections)If the scope of the ticket is not present in the existing token the following warning is thrown - 
 `WARNING b'{"status":"error","message":"This app hasn\'t been granted all required scopes to make this call. Read more about required scopes here: https://developers.hubspot.com/scopes.","correlationId":"4f0adda3-0fb9-4cb6-8d78-60b604f01259","errors":[{"message":"One or more of the following scopes are required.","context":{"requiredScopes":["tickets"]}}],"links":{"scopes":"https://developers.hubspot.com/scopes"},"category":"MISSING_SCOPES"}'
`

- As discussed with product, If a user selects a huge number of properties while making a field selection the tap execution is stopped with the following error message - 
`CRITICAL For stream - tickets, please select less number of properties as hubspot system doesn't allow huge request uri.`
`CRITICAL b'<html>\r\n<head><title>414 Request-URI Too Large</title></head>\r\n<body>\r\n<center><h1>414 Request-URI Too Large</h1></center>\r\n<hr><center>cloudflare</center>\r\n</body>\r\n</html>\r\n'`

# Manual QA steps
 - Tested the sync locally and tested on internal alpha (100608 | 383098)
 - Tested different scenarios like no access to tickets scope, 414 Request-URI Too Large

 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tap_hubspot/tests/unittests/test_tickets.py
```diff
@@ -1,117 +1,154 @@
...
-                    'id': '101',
-                    'type': 'ticket_to_contact'
-                }]
-            }
-        },
-    }]
-}
+class MockContext:
+    def get_catalog_from_id(self, stream_name):
+        return {"stream": "tickets", "tap_stream_id": "tickets",
```
rdeshmukh15, line 44: formating of the return value can be better

tap_hubspot/__init__.py
```diff
@@ -182,10 +199,16 @@ def get_field_schema(field_type, extras=False):
...
-    return {
-        field['name']: get_field_schema(field['type'], entity_name != 'contacts')
-        for field in data
-    }
+    if entity_name == "tickets":
+        return {
+            field['name']: get_field_type_schema(field['type'])
+            for field in data["results"]
+        }
+    else:
```
shantanu73, line 207: No need to put return statement in else block.
Do it like this :-
```
if condition:
    return value1
return value2
```

tap_hubspot/__init__.py
```diff
@@ -326,6 +352,8 @@ def request(url, params=None):
         timer.tags[metrics.Tag.http_status_code] = resp.status_code
         if resp.status_code == 403:
```
shantanu73, line 353: This block can also be modified as :-
```
if cond1:
   raise Exception1()
if cond2:
   raise Exception2()
resp.raise_for_status()
```

tap_hubspot/__init__.py
```diff
@@ -1112,7 +1138,10 @@ def do_sync(STATE, catalog):
         except SourceUnavailableException as ex:
             error_message = str(ex).replace(CONFIG['access_token'], 10 * '*')
             LOGGER.error(error_message)
-
+        except UriTooLongException as ex:
+            LOGGER.fatal(f"For stream - {stream.tap_stream_id}, please select less number of properties as "
+                         f"hubspot system doesn't allow huge request uri.")
+            raise ex
```
kethan1122, line 1144: Is there a specific number like Maximum number of fields that can be selected? If Yes, we can mention that number is log message.
`please select 150 or less number of fields`

tap_hubspot/tests/unittests/test_tickets.py
sgandhi1311: done

tap_hubspot/__init__.py
sgandhi1311: done

tap_hubspot/__init__.py
sgandhi1311: We don't have a specific number of fields. It totally depends on the character length.

tap_hubspot/__init__.py
```diff
@@ -139,6 +143,19 @@ def clean_state(state):
             LOGGER.info("%s - Removing last_sync_duration from state.", stream)
             state["bookmarks"][stream].pop("last_sync_duration", None)
 
+def get_selected_property_fields(catalog, mdata):
+
+    # mdata = metadata.to_map(catalog.get_stream(stream_name).metadata)
```
leslievandemark, line 148: can this comment be removed?

tap_hubspot/__init__.py
sgandhi1311: Done

tap_hubspot/__init__.py
sgandhi1311: removed

leslievandemark (approved):

kethan1122 (approved):

shantanu73 (approved):

PR #227, opened 2023-04-25
Fix bookmarking late record updates during sync

RushiT0122:
# Description of change
To handle records updated between start of the table sync and the end, store the current sync start in the state and not move the bookmark past this value.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/tests/test_deals.py
+++ b/tap_hubspot/tests/test_deals.py
@@ -1,10 +1,6 @@
 from tap_hubspot import sync_deals
 from unittest.mock import patch, ANY
 
-# from debugpy import listen, wait_for_client
-# listen(8000)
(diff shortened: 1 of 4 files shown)


tap_hubspot/tests/test_deals.py
```diff
@@ -1,14 +1,19 @@
 from tap_hubspot import sync_deals
 from unittest.mock import patch, ANY
 
+# from debugpy import listen, wait_for_client
+# listen(8000)
+# wait_for_client()
```
sgandhi1311, line 6: remove these debugger lines

sgandhi1311 (approved):

PR #232, opened 2023-07-19
start-date fix to handle data gaps

JYOTHINARAYANSETTY:
# Description of change
start-date fix to handle data gaps

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


bhtowles (commented):
This is not related to your change but I think while we have this open we can clean this second get_properties method up and remove the utc_today assignment since it is not being used for the static test

bhtowles (approved):

bhtowles (approved):

PR #235, opened 2023-09-15
Fix all fields, add missing fields

bhtowles:
# Description of change
Fix all fields, add missing fields

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tests/test_hubspot_all_fields.py
```diff
@@ -43,16 +43,24 @@ def get_matching_actual_record_by_pk(expected_primary_key_dict, actual_records):
 
 KNOWN_MISSING_FIELDS = {
     'contacts':{ # BUG https://jira.talendforge.org/browse/TDL-16016
-        'property_hs_latest_source_data_2',
         'property_hs_latest_source',
         'property_hs_latest_source_data_1',
-        'property_hs_timezone',
```
JYOTHINARAYANSETTY, line 49: Why are we deleting Fields from the list ?
Is it like we no longer have issue with these fields ?

JYOTHINARAYANSETTY (approved):

PR #240, opened 2023-11-01
Log the time difference of create time and post time

bhuvana-talend:
# Description of change
https://jira.talendforge.org/browse/TDL-20939
The change is to add more logging to decide on further steps to a discrepancy in bookmark test that occurs occassionally in stream contacts.
Log the time difference between the post message and the create time when a record is created to decide if we want to have a lookback window.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -1,6 +1,7 @@
 import datetime
 import random
 import uuid
+import sys
 
 import backoff
 import requests
@@ -16,6 +17,7 @@ class TestClient():
     V3_DEALS_PROPERTY_PREFIXES = {'hs_date_entered', 'hs_date_exited', 'hs_time_in'}
     BOOKMARK_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
     record_create_times = {}
+    time_difference=[]
 
     ##########################################################################
     ### CORE METHODS
@@ -759,7 +761,9 @@ class TestClient():
         elif stream == 'workflows':
             return self.create_workflows()
         elif stream == 'contacts':
-            return self.create_contacts()
+            records =  self.create_contacts()
+            self.record_create_times[stream]=self.time_difference
+            return records
         elif stream == 'deal_pipelines':
             return self.create_deal_pipelines()
         elif stream == 'email_events':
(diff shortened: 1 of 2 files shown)


tests/client.py
```diff
@@ -822,6 +823,10 @@ def create_contacts(self):
             ]
         }
 
+        #Get the current time in seconds
+        date= datetime.datetime.utcnow() - datetime.datetime(1970, 1, 1)
+        seconds =(date.total_seconds())
```
HarrisonMarcRose, line 828:
```suggestion     
        # Get the current time in seconds
        date = datetime.datetime.utcnow()
        seconds = datetime.datetime.timestamp(date)
```
This would only give you precision to the second. We probably want to go to millisecond or microsecond precision as this suggestion would do

tests/client.py
```diff
@@ -15,6 +15,7 @@ class TestClient():
     START_DATE_FORMAT = "%Y-%m-%dT00:00:00Z"
     V3_DEALS_PROPERTY_PREFIXES = {'hs_date_entered', 'hs_date_exited', 'hs_time_in'}
     BOOKMARK_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+    record_create_times = {}
```
HarrisonMarcRose, line 18: I don't see where we use this.  Can you point it out to me?

I see it now in the test. I was expecting it to be used in the client. See comment below.

tests/test_hubspot_bookmarks.py
```diff
@@ -48,24 +49,46 @@ def setUp(self):
...
+                    if stream in 'contacts':
+                        if self.test_client.time_difference < self.minimum:
+                            self.minimum = self.test_client.time_difference
+                        if self.test_client.time_difference > self.maximum:
+                            self.maximum = self.test_client.time_difference
+                        time_diff += self.test_client.time_difference 
+
+                if stream in 'contacts':
+                    self.avg = time_diff/5
+                    self.test_client.record_create_times[stream] = time_diff
```
HarrisonMarcRose, line 86: I don't think it is necessary to change the test to create more data.

It would be better to put all the data into your record_create_times in the client itself so that any test that creates this data by calling the client will have this data recorded.

Then once the test is finished you can use a tearDown method to analyze the data from the setup.  We could add this teardown to any test that uses the contacts stream to get data from multiple tests instead of just this one.

HarrisonMarcRose (changes requested):

tests/client.py
bhuvana-talend: Changed the time to milliseconds as suggested

tests/test_hubspot_bookmarks.py
bhuvana-talend: Moved the code to record the times to client. And, added teardown method to print the result at the end.

HarrisonMarcRose (approved):

PR #241, opened 2023-11-07
Tdl 24145

bhuvana-talend:
# Description of change
(https://jira.talendforge.org/browse/TDL-24145)
client.py - Added CRUD functionality for custom contact properties
test_hubspot_bookmarks.py - Call the create method in test data setup 

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -35,6 +35,14 @@ class TestClient():
                           jitter=None,
                           giveup=giveup,
                           interval=10)
+
+    def get_custom_property(self, url, params=dict()):
+        """Perform a GET using the standard requests method and logs the action"""
+        """Do not handle exception for get failure in custom contact properties"""
+        response = requests.get(url, params=params, headers=self.HEADERS)
+        LOGGER.info(f"TEST CLIENT | GET {url} params={params}  STATUS: {response.status_code}")
+        return response
+
     def get(self, url, params=dict()):
         """Perform a GET using the standard requests method and logs the action"""
         response = requests.get(url, params=params, headers=self.HEADERS)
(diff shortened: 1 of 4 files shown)


tests/client.py
```diff
@@ -773,6 +781,111 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
...
+              },
+              {
+                "label": "Option C",
+      "          value": "option_c"
+              }
+            ],
+            "formField": True
+        }
+
+        # generate a contacts record
```
bhtowles, line 877: Dev has recommended using raise_for_status() after requests.  As written we will just log bad responses and continue the test.  Is this the desired behavior?

tests/client.py
```diff
@@ -773,6 +781,111 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
...
+                "label": "Option A",
+                "value": "option_a"
+              },
+              {
+                "label": "Option B",
+                "value": "option_b"
+              },
+              {
+                "label": "Option C",
+      "          value": "option_c"
```
bhtowles, line 871: Extra white space

tests/test_hubspot_bookmarks.py
```diff
@@ -53,6 +53,9 @@ def create_test_data(self, expected_streams):
                                  for stream in expected_streams}
 
         for stream in expected_streams - {'contacts_by_company'}:
+            """ Create custom properties for contacts """
```
bhtowles, line 56: Is this meant to be a comment instead of a doc string?

tests/client.py
```diff
@@ -35,6 +35,14 @@ def giveup(exc):
                           jitter=None,
                           giveup=giveup,
                           interval=10)
+
+    def get_custom_property(self, url, params=dict()):
+        """Perform a GET using the standard requests method and logs the action"""
+        """Do not handle exception for get failure in custom contact properties"""
+        response = requests.get(url, params=params, headers=self.HEADERS)
```
bhtowles, line 42: Consider using raise_for_status()

tests/client.py
```diff
@@ -773,6 +781,111 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
...
+    def delete_custom_contacts_property(self):
+        """ Clearup the custom contact properties before CRUD test """
+        custom_properties = ['custom_string', 'custom_number', 'custom_date', 'custom_datetime', 'multi_pick']
+        for property in custom_properties:
+             get_url = f"{BASE_URL}/properties/v1/contacts/properties/named/" + property
+             response = self.get_custom_property(get_url)
+             if ( response.status_code == 404 ):
+                 continue
+             url = f"{BASE_URL}/properties/v1/contacts/properties/named/" + property
+             response = self.delete(url)
```
bhtowles, line 793: Consider using raise_for_status().  Should we re-try if we fail on a delete?

tests/test_hubspot_newfw_all_fields.py
```diff
@@ -0,0 +1,312 @@
...
+from base_hubspot import HubspotBaseCase
+from client import TestClient
+
+def get_matching_actual_record_by_pk(expected_primary_key_dict, actual_records):
+    ret_records = []
+    can_save = True
+    for record in actual_records:
+        for key, value in expected_primary_key_dict.items():
+            actual_value = record[key]
+            if actual_value != value:
```
bhtowles, line 13: Would like to run this and see in debugger.

tests/test_hubspot_newfw_all_fields.py
```diff
@@ -0,0 +1,312 @@
...
+                                actual_keys_adjusted.remove(key)
+
+                        self.assertSetEqual(expected_keys_adjusted, actual_keys_adjusted)
+
+    ##########################################################################
+    # Tests To Skip
+    ##########################################################################
+
+    @unittest.skip("Skip till all cards of missing fields are fixed. TDL-16145 ")
+    def test_all_fields_for_streams_are_replicated(self):
```
bhtowles, line 294: Seems like the same test is listed twice and skipped here the second time.  Confusing

tests/test_hubspot_newfw_all_fields.py
```diff
@@ -0,0 +1,312 @@
...
+        return "tt_hubspot_all_fields"
+
+    def streams_to_test(self):
+        """expected streams minus the streams not under test"""
+        return self.expected_streams().difference({
+            'owners',
+            'subscription_changes', # BUG_TDL-14938 https://jira.talendforge.org/browse/TDL-14938
+        })
+
+    def setUp(self):
```
bhtowles, line 174: Usually setUp is written to only run once even if multiple tests are called.  This implementation seems to only have one test so it may not be an issue but I'd like to maintain the pattern if possible.

tests/test_hubspot_newfw_all_fields.py
```diff
@@ -0,0 +1,312 @@
...
+            actual_value = record[key]
+            if actual_value != value:
+                can_save = False
+                break
+        if can_save:
+            ret_records.append(record)
+        can_save = True
+    return ret_records
+
+FIELDS_ADDED_BY_TAP = {
```
bhtowles, line 21: Many of these dicts, and methods will be useful in other tests.  I think most of them belong in base.

bhtowles (commented):
Will finish review tommorrow

tests/client.py
bhuvana-talend: post method is doing the raise_for_status check. create_contacts method is doing similar to what is beoing done here. I took that as the template.

tests/test_hubspot_bookmarks.py
bhuvana-talend: Meant to be a comment. I changed to #, don't really know the difference

tests/client.py
bhuvana-talend: This is intentional. I don't want to raise the exception - rather handle it. If the property is already there, I would delete it. See line# 789 where it is called. In the next line, I am checking the status

tests/client.py
bhuvana-talend: hmm... raise_for_status will log it as an error. I don't know if we want that.

tests/test_hubspot_newfw_all_fields.py
bhuvana-talend: Let me set up a time for this and to discuss the above comment

tests/test_hubspot_newfw_all_fields.py
bhuvana-talend: I have removed the testname and added the code as a comment to avoid confusion

tests/test_hubspot_newfw_all_fields.py
bhuvana-talend: Let's discuss on how to do this tap specific setUp

PR #242, opened 2023-11-08
Tdl 24434 custom object support

sgandhi1311:
# Description of change
Enhancing tap with support for custom CRM objects. Here's a summary of the key developments and considerations:

### Custom Object Handling
~~Tables for custom objects will be prefixed with "custom_" to distinguish them from standard objects in the backend. For example, if a customer has a custom object named "cars," the corresponding table in Stitch UI will be named "custom_cars."~~
To prevent confusion for customers, the custom object "cars" will be displayed simply as "cars" in the Stitch UI.

### Field Appearance
Stitch will support both default and custom properties. To streamline user selection, all properties will be raised to the top level. For instance, if a custom object "cars" includes a custom property "model," it will be displayed as "property_model."

### Associations
Expanded support for associations of custom objects with various objects, including 'emails, meetings, notes, tasks, calls, conversations, contacts, companies, deals, tickets.' Additionally, a support ticket has been raised with HubSpot to explore the feasibility of fetching associations between two custom objects.

# Manual QA steps
 - Verified the discovery of custom streams by retrieving custom object-related properties from the schema endpoint (/crm/v3/schemas).
 - Conducted local synchronization of records.
 - Updated the integration tests here - #243 
 
# Risks
 - The risk associated with this enhancement is deemed low, as we have implemented an exception handling mechanism for cases where the refresh token is missing the required scopes.
 
# Rollback steps
 - revert this branch


tap_hubspot/__init__.py
```diff
@@ -1104,6 +1112,85 @@ def sync_deal_pipelines(STATE, ctx):
...
+                raise RuntimeError(
+                    "Unexpected API response: {} not in {}".format(path, data.keys()))
+
+            for row in data[path]:
+                counter.increment()
+                yield row
+
+            if not data.get(more_key):
+                break
+            params['after'] = data.get(more_key).get('next').get('after')
```
luandy64, line 1133: Chains of `.get()` calls are not as safe as you think.

Consider the following scenarios. Which ones print and which ones error?

A.
```python
data = {}
more_key = "paging"

print(data.get(more_key).get('next').get('after'))
```

B.
```python
data = {"paging": {}}
more_key = "paging"

print(data.get(more_key).get('next').get('after'))
```

C.
```python
data = {"paging": {"next": {}}}
more_key = "paging"

print(data.get(more_key).get('next').get('after'))
```

luandy64: [spoilers.txt](https://github.com/singer-io/tap-hubspot/files/13337552/spoilers.txt)

luandy64: [spoilers](https://gist.github.com/luandy64/e0ba5d71ab6415190ff572f1c63e8369)

sgandhi1311: Attribute error for options A and B. Option C gives us the default value as None.

In this case, I think I can modify as below - 
```
params['after'] = data.get(more_key, {}).get('next', {}).get('after', None)
if params['after'] is None:
    break
```

Vi6hal (approved):

tap_hubspot/__init__.py
```diff
@@ -1131,6 +1220,70 @@ class Stream:
...
+                custom_schema_top_level = {'property_{}'.format(k): v for k, v in custom_schema.items()}
+                schema['properties'].update(custom_schema_top_level)
+
+                final_schema = singer.resolve_schema_references(schema, refs)
+                custom_schema_path = get_abs_path('schemas/{}.json'.format(stream_id))
+
+                # Write data to the JSON file
+                with open(custom_schema_path, 'w') as json_file:
+                    json.dump(final_schema, json_file)
+
```
Vi6hal, line 1261: This is a very unconventional approach for handling custom streams, 
This adds another set of disk-io operations, it may not have any significant performance degradation but we are writing these schemas from memory to disk only to be read again to memory.
can we change this implementation to something where these custom schemas might be returned to the discovery function and it processes it?

Vi6hal (changes requested):

tap_hubspot/__init__.py
sgandhi1311: Agreed, I have removed the disk IO operations. And stored the new custom schemas in the variable till the time of execution.

Vi6hal (approved):

tap_hubspot/__init__.py
```diff
@@ -1131,6 +1220,67 @@ class Stream:
...
+            warning_message = str(ex).replace(CONFIG['access_token'], 10 * '*')
+            LOGGER.warning(warning_message)
+        return custom_streams
+
+    elif mode == "SYNC":
+        custom_objects = [custom_object["name"] for custom_object in gen_request_custom_objects("custom_objects_schema", custom_objects_schema_url, {}, 'results', "paging")]
+        for stream in catalog["streams"]:
+            if stream["tap_stream_id"] in custom_objects:
+                STREAMS.append(Stream(stream["tap_stream_id"], sync_custom_object_records, ['id'], 'updatedAt', 'INCREMENTAL'))
+        return custom_objects
```
dsprayberry, line 1269: I don't have a great idea of what it would look like, but I feel like this could be made into two separate functions to keep discovery functionality inside the discover_schemas() function?

I suppose that might not be worth the effort given that this tap has everything inside init instead of separate sync / discover files, but I felt like it was worth thinking about at least

sgandhi1311: Yeah, I thought of extending the existing function - discover_schemas() but it was looking complex, therefore, I created a new function. Maybe I can put it in the backlog and take care while refactoring the code (main motive is to remove the redundant code.)

tap_hubspot/__init__.py
```diff
@@ -1104,6 +1112,87 @@ def sync_deal_pipelines(STATE, ctx):
...
+                counter.increment()
+                yield row
+
+            if not data.get(more_key):
+                break
+            params['after'] = data.get(more_key, {}).get('next', {}).get('after', None)
+            if params['after'] is None:
+                break
+
+def sync_records(stream_id, primary_key, bookmark_key, catalog, STATE, params, is_custom_object=False):
```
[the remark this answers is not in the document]
dsprayberry: I think that until that gets done, we should keep the name more explicit. We can always change it back when the refactor happens and then keep the function obvious in the event that the refactor gets delayed or deprioritized

tap_hubspot/__init__.py
dsprayberry: 👍

tap_hubspot/__init__.py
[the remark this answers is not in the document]
sgandhi1311: done

dsprayberry (approved):

dsprayberry (approved):

PR #244, opened 2023-11-17
Create custom contact properties and verify replication of those properties

bhuvana-talend:
# Description of change
(https://jira.talendforge.org/browse/TDL-24145)
client.py - Method to create custom contact properties and add it to the contacts
test_hubspot_bookmarks.py - Call the create method in test data setup

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -7,8 +7,8 @@ import backoff
 import requests
 from base import HubspotBaseTest
 from tap_tester import LOGGER
-from copy import deepcopy
 import time
+from copy import deepcopy
 
 DEBUG = False
 BASE_URL = "https://api.hubapi.com"
--- a/tests/client.py
+++ b/tests/client.py
@@ -8,6 +8,7 @@ import requests
 from base import HubspotBaseTest
 from tap_tester import LOGGER
 from copy import deepcopy
+import time
 
 DEBUG = False
 BASE_URL = "https://api.hubapi.com"
--- a/tests/client.py
+++ b/tests/client.py
@@ -805,6 +805,13 @@ class TestClient():
             return self.create_companies()
         elif stream == 'contact_lists':
             return self.create_contact_lists()
+        elif stream == 'static_contact_lists':
+            staticlist = self.create_contact_lists(dynamic=False)
+            listId = staticlist[0].get('listId')
+            records =  self.create('contacts')
+            contact_email =  records[0].get('properties').get('email').get('value')
+            self.add_contact_to_contact_list(listId, contact_email)
+            return staticlist
         elif stream == 'contacts_by_company':
             return self.create_contacts_by_company(company_ids, times=times)
         elif stream == 'engagements':
@@ -816,8 +823,9 @@ class TestClient():
         elif stream == 'workflows':
             return self.create_workflows()
         elif stream == 'contacts':
+            LOGGER.info("self.record_create_times is %s", self.record_create_times)
             if stream not in self.record_create_times.keys():
-                self.record_create_times[stream]=[]
+                self.record_create_times['contacts']=[]
             records =  self.create_contacts()
             return records
         elif stream == 'deal_pipelines':
@@ -1051,7 +1059,7 @@ class TestClient():
         records = [response]
         return records
 
-    def create_contact_lists(self):
+    def create_contact_lists(self, dynamic=True):
         """
         HubSpot API https://legacydocs.hubspot.com/docs/methods/lists/create_list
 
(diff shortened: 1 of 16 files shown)


bhtowles:
Where do these custom properties show up?  I would not expect them to necessarily change bookmarks as I assume it is an add on to an existing stream since there is no update in base or streams_to_test.  Should we move this to the all fields test and actually test that the custom data is being replicated?

tests/client.py
```diff
@@ -778,6 +780,103 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
...
+            "formField": True
+        }
+        data.append(deepcopy(property))
+        # generate a contacts record
+
+        for current_data in data:
+            try:
+                response = self.post(url, current_data)
+                LOGGER.info("response is %s", response)
+            except Exception as e:
```
HarrisonMarcRose, line 875: This Exception is too broad.  We should be a specific with exceptions as possible as this will swallow any exception, some not intended here.

Also, I'm not sure why you would get an exception if data already exists instead of getting a 4xx response code?

tests/client.py
```diff
@@ -778,6 +780,103 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
...
+                "value": "option_b"
+              },
+              {
+                "label": "Option C",
+                "value": "option_c"
+              }
+            ],
+            "formField": True
+        }
+        data.append(deepcopy(property))
```
HarrisonMarcRose, line 868: It is great that you are creating a variety of data types in the custom fields.  Since you went through the pain of creating them we should also use them all when creating a contact - instead of just the first two - to make sure that we support all of the types.

HarrisonMarcRose (commented):
See comments

tests/client.py
bhuvana-talend: yeah, it is a 409, which was causing the test to fail. I'll handle it specifically instead of a broader one.

tests/client.py
bhuvana-talend: :) was expecting this comment.. The other types seems complicated, I'll need more time to figure out how to do. Can I write a separate card for that ?

HarrisonMarcRose: sure.

tests/client.py
HarrisonMarcRose: I think we miscommunicated.  I meant in the `except Exception as e` we should actually change the word `Exception` to a more specific exception.  I don't know what kind it is but based on the 409 maybe it is an RequestException or HTTPError?  However, I didn't think a 409 would actually raise an exception.  If you are trying to do this until you get a successful status code maybe another way would be better. Let's talk about what you are trying to achieve to determine the best way to go about it.

tests/client.py
```diff
@@ -843,6 +945,14 @@ def create_contacts(self):
...
         data = {
             "properties": [
+                {
+                    "property": "custom_string",
+                    "value": "custom_string_value"
+                },
+                {
+                    "property": "custom_number",
+                    "value": 1567
+                },
```
HarrisonMarcRose, line 955: Since we are setting custom properties for every call of `create_contacts` we need the `create_custom_contact_properties` to be called prior to calling the `create_contacts`.  I see you are doing this in the test, but what about other tests that would need to add contacts, those would probably break if the custom properties were not present. I believe we should call `create_custom_contact_properties` in the `__init__` of the client so that we call it once when it starts up instead of depending on the tester to know about this or remember to add it in each test.

HarrisonMarcRose (commented):
Lets talk about these comments

tests/client.py
bhuvana-talend: the post call returns a response code of 409 and in post method the exception is raised. That's why I am handling it here. I can change the exception to a specific one to make it more readable.

tests/client.py
bhuvana-talend: So, creating the custom contact property is a one time task. After my first successful run of the test, it never creates and gives a 409, that's why I am skipping it. Probably doing it in init is a good idea. Will try that

HarrisonMarcRose: I think you moved it to setup. I meant the init of the client so anyone using the client makes sure it is in good shape, vs just for your test.

tests/client.py
```diff
@@ -833,6 +835,107 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
...
+            try:
+                response = self.post(url, current_data)
+                LOGGER.info("response is %s", response)
+            # Setting up the property is a one time task, If exception occurs because, it already exists, ignore
+            except Exception as DataAlreadyExistsExcp:
+                LOGGER.info("Data already exists for %s", current_data)
+                if '409' in str(DataAlreadyExistsExcp):
+                    pass
+                else:
+                    response.raise_for_status()
```
HarrisonMarcRose, line 936: I don't think I"m communicating well enough.  doing `except Exception` is a no-no as this is a very broad exception.  We need to figure out the specific exception that is being thrown and catch that.  To determine what exception is being thrown you can remove the try/except and run the test to see what exception is raised. Catch that one.

HarrisonMarcRose: > [Exception](https://docs.python.org/3/library/exceptions.html#Exception) can be used as a wildcard that catches (almost) everything. However, it is good practice to be as specific as possible with the types of exceptions that we intend to handle, and to allow any unexpected exceptions to propagate on.

https://docs.python.org/3/tutorial/errors.html

bhuvana-talend: Sorry Harrison - I think I got it right now :) Now handling it as requests.exceptions.HTTPError

HarrisonMarcRose (approved):

PR #245, opened 2023-12-06
test to validate the replication of static contact list - Tdl 18267

bhuvana-talend:
h
ttps://jira.talendforge.org/browse/TDL-18267
test_hubspot_bookmarks.py
- Modified the first list of the contact list to be static
- Update the last item in the initial list intead of first
client.py
- modify create_contact_list to accept a parameter for dynamic ( true or false )
- add a method add_contact_to_contact_list to assign a contact to the static list
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -7,6 +7,7 @@ import backoff
 import requests
 from base import HubspotBaseTest
 from tap_tester import LOGGER
+import time 
 
 DEBUG = False
 BASE_URL = "https://api.hubapi.com"
@@ -751,9 +752,12 @@ class TestClient():
         elif stream == 'contact_lists':
             return self.create_contact_lists()
         elif stream == 'static_contact_lists':
-            static_list = self.create_contact_lists(False)
-            #records =  self.create_contacts()
-            #record = self.add_contact_to_contact_list(list_id, contact_email)
+            staticlist = self.create_contact_lists(False)
+            if stream not in self.record_create_times.keys():
+                self.record_create_times['contacts']=[]
+            records =  self.create_contacts()
+            self.add_contact_to_contact_list()
+            return staticlist
         elif stream == 'contacts_by_company':
             return self.create_contacts_by_company(company_ids, times=times)
         elif stream == 'engagements':
@@ -765,8 +769,9 @@ class TestClient():
         elif stream == 'workflows':
             return self.create_workflows()
         elif stream == 'contacts':
+            LOGGER.info("self.record_create_times is %s", self.record_create_times)
             if stream not in self.record_create_times.keys():
-                self.record_create_times[stream]=[]
+                self.record_create_times['contacts']=[]
             records =  self.create_contacts()
             return records
         elif stream == 'deal_pipelines':
@@ -789,14 +794,14 @@ class TestClient():
         Hubspot API https://legacydocs.hubspot.com/docs/methods/contacts/create_contact
         """
         record_uuid = str(uuid.uuid4()).replace('-', '')
-        value = f"{record_uuid}@stitchdata.com"
+        self.contact_email = f"{record_uuid}@stitchdata.com"
 
         url = f"{BASE_URL}/contacts/v1/contact"
         data = {
             "properties": [
                 {
                     "property": "email",
-                    "value": value
+                    "value": self.contact_email
                 },
                 {
                     "property": "firstname",
@@ -858,7 +863,7 @@ class TestClient():
         records = self.denest_properties('contacts', [get_resp])
 
         # Add the email to add it to the contact list
-        self.contacts_added.add(value)
+        self.contacts_added.append(self.contact_email)
 
         return records
 
@@ -919,10 +924,12 @@ class TestClient():
         }
         # generate a record
         response = self.post(url, data)
+        self.list_id = response.get("listId")
         records = [response]
+        LOGGER.info("dynamic contact list is %s", records)
         return records
 
-    def add_contact_to_contact_list(self, list_id, contact_email):
+    def add_contact_to_contact_list(self):
         """
         HubSpot API https://legacydocs.hubspot.com/docs/methods/lists/create_list
 
@@ -933,15 +940,17 @@ class TestClient():
         record_uuid = str(uuid.uuid4()).replace('-', '')
         value = f"@hubspot{record_uuid}"
 
-        url = f"{BASE_URL}/contacts/v1/lists/{list_id}/add"
+        url = f"{BASE_URL}/contacts/v1/lists/{self.list_id}/add"
         data = {
             "emails": [
-               contact_email
+               self.contact_email
             ]
         }
         # generate a record
+        LOGGER.info("Post URL is %s", url)
         response = self.post(url, data)
         records = [response]
+        LOGGER.info("updated contact_list is %s", records)
         return records
 
     def create_contacts_by_company(self, company_ids=[], contact_records=[], times=1):
--- a/tests/client.py
+++ b/tests/client.py
@@ -17,6 +17,7 @@ class TestClient():
     V3_DEALS_PROPERTY_PREFIXES = {'hs_date_entered', 'hs_date_exited', 'hs_time_in'}
     BOOKMARK_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
     record_create_times = {}
+    contacts_added = []
 
     ##########################################################################
     ### CORE METHODS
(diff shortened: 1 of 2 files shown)


tests/client.py
```diff
@@ -17,6 +18,7 @@ class TestClient():
     V3_DEALS_PROPERTY_PREFIXES = {'hs_date_entered', 'hs_date_exited', 'hs_time_in'}
     BOOKMARK_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
     record_create_times = {}
+    contacts_added = []
```
HarrisonMarcRose, line 21: see comment below

tests/client.py
```diff
@@ -893,23 +907,50 @@ def create_contact_lists(self):
...
-                    "value": f"@hubspot{record_uuid}",
+                    "value": value,
                     "property": "twitterhandle",
                     "type": "string"
                 }]
             ]
         }
         # generate a record
         response = self.post(url, data)
+        self.list_id = response.get("listId")
```
HarrisonMarcRose, line 927: The list_id is in the response you are returning, there is no need to store it separately

tests/client.py
```diff
@@ -749,6 +751,13 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
             return self.create_companies()
         elif stream == 'contact_lists':
             return self.create_contact_lists()
+        elif stream == 'static_contact_lists':
+            staticlist = self.create_contact_lists(False)
+            if stream not in self.record_create_times.keys():
+                self.record_create_times['contacts']=[]
+            records =  self.create_contacts()
+            self.add_contact_to_contact_list()
```
HarrisonMarcRose, line 759: We should get the list_id out of the `staticlist` above and send it along here for the method to have.  I think the same is true for the contact_email in the created contact.

tests/test_hubspot_bookmarks.py
```diff
@@ -30,11 +30,12 @@ def name():
...
     def streams_to_test(self):
         """expected streams minus the streams not under test"""
-        expected_streams = self.expected_streams().difference(STREAMS_WITHOUT_CREATES)
+        """expected_streams = self.expected_streams().difference(STREAMS_WITHOUT_CREATES)
 
         return expected_streams.difference({
             'subscription_changes', # BUG_TDL-14938 https://jira.talendforge.org/browse/TDL-14938
-        })
+        })"""
+        return {'contact_lists'}
```
HarrisonMarcRose, line 38: I think this was just during testing to only test contact_lists but shouldn't be in the final version

tests/test_hubspot_bookmarks.py
```diff
@@ -116,11 +123,13 @@ def test_run(self):
...
             self.expected_records['contacts_by_company'] += record
 
-
         # Update 1 record from the test seutp for each stream that has an update endpoint
+        # Update the last record that was initially created, instead of first record
+        # - Workaround for https://jira.talendforge.org/browse/TDL-24615
         for stream in expected_streams - STREAMS_WITHOUT_UPDATES:
             primary_key = list(self.expected_primary_keys()[stream])[0]
-            record_id = self.expected_records[stream][0][primary_key]
+            index_to_update = len(self.expected_records[stream])-1
```
HarrisonMarcRose, line 131: This workaround will affect all streams and not just the contact lists.  And the last record created will be synced regardless of it is updated or not so this isn't testing this for other streams. I think a better workaround would be to add this stream to STREAMS_WITHOUT_UPDATES so this loop just doesn't run for contact lists until it works.

HarrisonMarcRose (changes requested):

tests/client.py
```diff
@@ -851,6 +862,9 @@ def create_contacts(self):
         get_resp['versionTimestamp'] = converted_versionTimestamp
         records = self.denest_properties('contacts', [get_resp])
 
+        # Add the email to add it to the contact list
+        self.contacts_added.append(self.contact_email)
+
```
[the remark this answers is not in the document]
bhuvana-talend: Removed this variable as it is not used anywhere

tests/client.py
```diff
@@ -749,6 +751,13 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
             return self.create_companies()
         elif stream == 'contact_lists':
             return self.create_contact_lists()
+        elif stream == 'static_contact_lists':
+            staticlist = self.create_contact_lists(False)
+            if stream not in self.record_create_times.keys():
+                self.record_create_times['contacts']=[]
+            records =  self.create_contacts()
```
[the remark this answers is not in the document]
bhuvana-talend: Modified the code as per suggestion

tests/client.py
bhuvana-talend: Done. Making use of the list_id from response

tests/test_hubspot_bookmarks.py
bhuvana-talend: Removed

tests/test_hubspot_bookmarks.py
bhuvana-talend: Changed it to be the first record as before.

tests/client.py
```diff
@@ -802,6 +803,13 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
             return self.create_companies()
         elif stream == 'contact_lists':
             return self.create_contact_lists()
+        elif stream == 'static_contact_lists':
+            staticlist = self.create_contact_lists(False)
```
HarrisonMarcRose, line 807:
```suggestion
            staticlist = self.create_contact_lists(dynamic=False)
```

tests/test_hubspot_bookmarks.py
```diff
@@ -57,6 +57,8 @@ def create_test_data(self, expected_streams):
         for stream in expected_streams - {'contacts_by_company'}:
             if stream == 'contacts': 
                 self.times=10
+            elif stream == 'contact_lists':
+                self.times=2
```
HarrisonMarcRose, line 61: I'm not sure why we are changing the number of times we are running this, but IIRC we need at least three so that we can add 3 on the first sync and then remove one record and update one record which wouldn't be synced on the next sync.  And the last item added is usually resynced on the next sync. So you need 3 records.  I'm curious if you meant this to be static_contact_lists since we create an extra one of those?

tests/test_hubspot_bookmarks.py
```diff
@@ -65,9 +67,13 @@ def create_test_data(self, expected_streams):
                 self.expected_records['email_events'] += email_records
             else:
                 # create records, one will be updated between syncs
+                # create one static list and the rest dynamic list
                 for _ in range(self.times):
                     record = self.test_client.create(stream)
                     self.expected_records[stream] += record
+                if stream in 'contact_lists':
```
HarrisonMarcRose, line 74: I'm not sure this is doing what you expect 
```suggestion
                if stream == 'contact_lists':
```
because you are comparing stream to a string with `in` it will only be true if stream is one of the characters in the string.  so if stream = `c` it would be true.  In is generally used with a list or tuple.

HarrisonMarcRose (commented):
Lets talk this over when you are ready

tests/test_hubspot_bookmarks.py
bhuvana-talend: yes, that is the idea. I want one of the 3 to be a static_contact_list

tests/client.py
bhuvana-talend: Done

tests/client.py
```diff
@@ -813,8 +821,9 @@ def create(self, stream, company_ids=[], subscriptions=[], times=1):
         elif stream == 'workflows':
             return self.create_workflows()
         elif stream == 'contacts':
+            LOGGER.info("self.record_create_times is %s", self.record_create_times)
```
HarrisonMarcRose, line 824: do we need this logger still?

tests/client.py
```diff
@@ -965,6 +975,31 @@ def create_contact_lists(self):
...
+        value = f"@hubspot{record_uuid}"
+
+        url = f"{BASE_URL}/contacts/v1/lists/{list_id}/add"
+        data = {
+            "emails": [
+               contact_email
+            ]
+        }
+        # generate a record
+        LOGGER.info("Post URL is %s", url)
```
HarrisonMarcRose, line 999: Is this to get the list id in the logs?

HarrisonMarcRose (approved):
Looks great. Not sure if we need some of the logging but all good.

bhuvana-talend:
merging

PR #250, opened 2024-01-19
Tdl 24590 fix contacts to company

RushiT0122:
# Description of change
Currently we make a request to retrieve `contacts_by_company` per `company_id`. But this results in performance issue [TDL-24590](https://jira.talendforge.org/browse/TDL-24590), if we have large number of company ids. To overcome this issue, we will make one request for 250 company ids in a batch to retrieve `contacts_by_company` which will reduce number of `contacts_by_company` requests almost by 250 times.

# Manual QA steps
 - Verified number of records match with original implementation
 - Verified different interrupted sync scenarios
     - No offset set for `contacts_by_company` to simulate existing connection scenario
     - `contacts_by_company` offset is lesser than `companies` offset
     - `contacts_by_company` offset is greater than `companies` offset (observed this scenario in internal testing)
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -77,6 +77,7 @@ ENDPOINTS = {
     "companies_recent":     "/companies/v2/companies/recent/modified",
     "companies_detail":     "/companies/v2/companies/{company_id}",
     "contacts_by_company":  "/companies/v2/companies/{company_id}/vids",
+    "contacts_by_company_v3_batch_read": "/crm/v3/associations/company/contact/batch/read",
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.13.1
```
rdeshmukh15, line 3: in setup.py version mentioned is 2.14.0 and here it is 2.13.1
These should be in sync

rdeshmukh15 (approved):

PR #263, opened 2024-09-12
Tdl 26414 rename custom objects

RushiT0122:
# Description of change
This PR addresses an issue where `custom objects` in HubSpot can have the same names as `standard objects` (e.g., contacts, campaigns), leading to extraction failures [TDL-26380](https://jira.talendforge.org/browse/TDL-26380). 

The fix involves renaming custom objects that share names with standard objects to a distinct `custom_object_<standard_object_name>`. This ensures that custom objects are clearly differentiated from standard objects, preventing conflicts and extraction issues.

# Manual QA steps

- [x] Verify that streams of conflicting custom objects have been renamed correctly.
- [x] Verify that custom objects initially named custom_object_<standard_object_name> are extracted as expected.
- [x] Verify that the standard and renamed custom object streams are extracted properly.

 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -1155,7 +1155,7 @@ def sync_custom_objects(stream_id, primary_key, bookmark_key, catalog, STATE, pa
     """
     mdata = metadata.to_map(catalog.get('metadata'))
     if is_custom_object:
-        url = get_url("custom_objects", object_name=catalog["stream_alias"])
+        url = get_url("custom_objects", object_name=catalog["table_name"])
     else:
         url = get_url(stream_id)
     max_bk_value = bookmark_value = utils.strptime_with_tz(
@@ -1180,7 +1180,7 @@ def sync_custom_objects(stream_id, primary_key, bookmark_key, catalog, STATE, pa
                 # transforms the data and filters out the selected fields from the catalog
                 record = transformer.transform(lift_properties_and_versions(row), schema, mdata)
                 singer.write_record(stream_id, record, catalog.get(
-                    'stream_alias'), time_extracted=utils.now())
+                    'stream'), time_extracted=utils.now())
             if modified_time and modified_time >= max_bk_value:
                 max_bk_value = modified_time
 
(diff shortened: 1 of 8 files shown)


sgandhi1311:
Can you please pull the latest changes from the master branch?

tap_hubspot/__init__.py
```diff
@@ -1422,6 +1426,7 @@ def discover_schemas():
         LOGGER.info('Loading schema for Custom Object - %s', custom_stream["stream"].tap_stream_id)
         result['streams'].append({'stream': custom_stream["stream"].tap_stream_id,
                                   'tap_stream_id': custom_stream["stream"].tap_stream_id,
+                                  "table_name": custom_stream["custom_object_name"],
```
sgandhi1311, line 1429: Is it necessary to have new key `table_name`? Because it looks like standard ojects won't be having that key.

tests/base.py
```diff
@@ -170,6 +170,20 @@ def expected_metadata(self):  # DOCS_BUG https://stitchdata.atlassian.net/browse
...
                 self.OBEYS_START_DATE: True
+            },
+            "custom_object_campaigns": {
+                self.PRIMARY_KEYS: {"id"},
+                self.REPLICATION_METHOD: self.INCREMENTAL,
+                self.REPLICATION_KEYS: {"updatedAt"},
+                self.EXPECTED_PAGE_SIZE: 100,
+                self.OBEYS_START_DATE: True
+            },
+            "custom_object_contacts": {
```
sgandhi1311, line 181: How about testing the corner case scenario for the stream name like - `custom_object_custom_object_contacts`?

tap_hubspot/__init__.py
RushiT0122: Yes, this `table_name` property value will be reflected on the app and in the extracted records. Whereas `stream` will retain the original object name which we need to make API requests.

tests/base.py
RushiT0122: We only rename custom objects stream if their original name matches with standard stream names like `contacts`, `campaigns`, `deals`. Therefore, if custom object is named as `custom_object_custom_object_contacts` then we won't take any action on it.

sgandhi1311 (approved):

PR #270, opened 2025-07-23
Qa/fix hub automatic

nk7up:
# Description of change
Fix to handle the problematic unsynced streams. These will updated once the results are consistent.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -221,6 +221,10 @@ class HubspotBaseTest(BaseCase):
     def expected_streams(self):
         """A set of expected stream names"""
         return set(self.expected_metadata().keys())
+    
+    def unsynced_streams(self):
+        """A set of streams that are known to have sync issues and should be skipped in assertion errors."""
+        return {"subscription_changes", "email_events"} 
 
     def expected_replication_keys(self):
         """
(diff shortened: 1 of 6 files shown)


tests/base.py
```diff
@@ -221,6 +221,24 @@ def expected_replication_method(self):
...
+        """A set of streams that are known to have sync issues and should be skipped in assertion errors."""
+        return {"subscription_changes", "email_events"}
+    
+    def validate_failed_sync_streams(self, stream, *synced_data_records):
+        """
+        Validate if the stream is in the known failed sync streams.
+        If it is, log a warning and skip the assertion.
+        If not, raise an error indicating the stream is missing from synced_records.
+        """
+        if stream in self.failed_sync_streams():
```
bhtowles, line 235: I thought we wanted to add a second "if" here to see if the failed sync stream has records or not.  If it does have records we can log that, return True, and test it.  If it does not have records (as we expect) then we can "continue" to skip.  I'm assuming adding that e-mail did not fix the email_events stream.  Can you reply to the email or do any actions to the e-mail in the UI?  Maybe that would count as an email event?

bhtowles (approved):

PR #273, opened 2025-08-06
SAC-28363: Adding co_firsts stream to streams_to_test to handle the inconsistencies

nk7up:
# Description of change
Handling the inconsistent streams "co_firsts" since they do not get processed quickly

Issue: https://qlik-dev.atlassian.net/browse/SAC-28363

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


bhtowles (approved):
Commit message is misleading as we are adding co_firsts to the set of streams to subtract (difference) from streams to test to exclude it from the pagination test.

PR #301, opened 2026-05-19
 Migrate engagements sync from /paged to /modified/after cursor-based pagination

surya-hubs:
*Submitted on behalf of HubSpot Engineering *

# Description of change
Migrate `sync_engagements()` from `GET /engagements/v1/engagements/paged` to the new `GET /engagements/v1/engagements/modified/after` cursor-based endpoint.

The old `/paged` path does full offset scans with client-side `lastUpdated` filtering, generating significant egress ). r

Changes:
- Bootstrap: converts existing `lastUpdated` bookmark to initial `after` timestamp
- Durable state: persists opaque `cursor` bookmark; clears stale `lastUpdated` on completion
- Crash recovery: in-flight `offset.cursor` takes precedence over `cursor` for interrupted runs
- Removed `current_sync_start` 

# Rollback steps
 - revert this branch

#### AI generated code
https://internal.qlik.dev/general/ways-of-working/code-reviews/#guidelines-for-ai-generated-code
- [x] this PR has been written with the help of GitHub Copilot or another generative AI tool

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 4.4.0
+  * Migrate engagements sync from `/paged` to `/engagements/v1/engagements/modified/after` with opaque cursor-based pagination for incremental, crash-safe resume.
+
 ## 4.3.0
   * Always write bookmarks for form_submissions and list_memberships streams, even when there are no records. [#292](https://github.com/singer-io/tap-hubspot/pull/292)
 
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 from setuptools import setup
 
 setup(name='tap-hubspot',
-      version='4.3.0',
+      version='4.4.0',
       description='Singer.io tap for extracting data from the HubSpot API',
       author='Stitch',
       url='http://singer.io',
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -144,18 +144,17 @@ def write_current_sync_start(state, tap_stream_id, start):
     return singer.write_bookmark(state, tap_stream_id, "current_sync_start", value)
 
 def get_engagements_cursor(state, start):
-    offset = singer.get_offset(state, 'engagements') or {}
-    if offset.get('cursor'):
-        return offset['cursor']
-
+    # Continue from the opaque cursor written after a successful modified/after sync.
     cursor = singer.get_bookmark(state, 'engagements', 'cursor')
     if cursor:
         return cursor
 
+    # Support older state that stored the modified/after cursor as the after bookmark.
     previous_after = singer.get_bookmark(state, 'engagements', 'after')
     if previous_after:
         return previous_after
 
+    # Migrate timestamp-based state by seeding modified/after with epoch millis.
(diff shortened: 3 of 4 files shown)


sgandhi1311:
@surya-hubs can you please fix the above review comments?

surya-hubs:
@sgandhi1311  done. please take a look. Thank you!

sgandhi1311:
@surya-hubs  I tested your changes locally and noticed a difference in the final state produced after sync.

**Your branch:**
```json
{"type": "STATE", "value": {"currently_syncing": null, "bookmarks": {"engagements": {"cursor": "cursor:MTc4MTgyNzIwMDAwMAoxNzgxODc0ODMzNjE2CjMuNzc1NjAzNjQ3NzZFMTE"}}}}
```

**`master` branch:**
```json
{"type": "STATE", "value": {"currently_syncing": "engagements", "bookmarks": {"engagements": {"current_sync_start": null, "lastUpdated": "2026-06-19T13:13:53.616000Z"}}}}`
```

The `lastUpdated` bookmark key is absent in your branch's state. Was this intentional? Since the `engagements` stream uses an **INCREMENTAL** replication strategy, the replication key must be preserved in state for subsequent syncs to work correctly. Removing or replacing it would be a breaking change and would require a **major version bump** — something I'd prefer to avoid if possible.

Could you also share the HubSpot documentation or changelog that motivated this change?

surya-hubs:
@sgandhi1311  


> Could you also share the HubSpot documentation or changelog that motivated this change?


**Docs that motivated the change:** the endpoint is publicly documented here — [Get engagements modified after](https://developers.hubspot.com/docs/api-reference/legacy/crm/activities/engagements/get-engagements-v1-engagements-modified-after). It's cursor-paginated with no 10k offset cap, and its `after` parameter accepts **either** a Unix-ms timestamp **or** a cursor token — which is exactly what makes the bootstrap from the old timestamp bookmark seamless.


> The lastUpdated bookmark key is absent in your branch's state. Was this intentional? Since the engagements stream uses an INCREMENTAL replication strategy, the replication key must be preserved in state for subsequent syncs to work correctly. Removing or replacing it would be a breaking change and would require a major version bump — something I'd prefer to avoid if possible


Thanks for testing this locally and catching the state difference — good catch.

Yeah the change was intentional.  However, the change is transparent to the system since the code can now read state with both `lastUpdated`( to preserve backward compatibility)  and opaque `cursor` present in the state.  When it writes back , however,  it writes the opaque cursor.  Subsequent syncs read the cursor after the update. 
All this should be completely transparent to clients updating the library and would be a non breaking change. 

Let me know if i missed something here .

sgandhi1311:
This PR looks great. Thank you for the contribution, @surya-hubs !
Before I can approve and merge into master, could you resolve the conflicts in CHANGELOG.md?

surya-hubs:
Thank you @sgandhi1311 !! 

Fixed merge conflicts.

sgandhi1311 (approved):

sgandhi1311:
@surya-hubs ~~Please merge my integration tests in your branch.
PR - https://github.com/surya-hubs/tap-hubspot/pull/1~~

PR #302, opened 2026-05-29
Exclude unauthorized streams in catalog

MuralidharT03:
# Description of change

This PR adds a discovery-time permission probe to HubSpot stream endpoints so the generated Singer catalog excludes streams that the provided credentials cannot read (403), and fails discovery when **no** parent streams are readable.

**Changes:**
- Add `STREAM_ACCESS_ENDPOINTS` — a single source of truth mapping each parent stream to its probe endpoint, HTTP method, and minimal params for access checking
- Add `check_stream_access(stream_name)` — probes a stream via a minimal GET (or POST for `contact_lists`) request and returns `True`/`False` on 403 Forbidden; child streams always return `True` since access is governed by their parent
- Add `_get_accessible_streams(streams)` — filters the full stream list to only accessible ones, prunes child streams whose parent was excluded, logs warnings for excluded streams, and raises `HubspotForbiddenError` when all parent streams are forbidden
- Update `discover_schemas()` to call `_get_accessible_streams()` before loading schemas, so only authorized streams appear in the generated catalog
- Add unit tests covering access probing (GET and POST paths), stream filtering behavior, child stream pruning, all-streams-blocked error case, and `STREAM_ACCESS_ENDPOINTS` completeness validation

Jira: [SAC-30970](https://qlik-dev.atlassian.net/browse/SAC-30970)

# Manual QA steps
- [x] Discovery: `tap-hubspot --config config.json --discover` — verify catalog contains only accessible streams
- [x] Discovery with blocked credentials — verify `HubspotForbiddenError` is raised when no parent streams are accessible
- [x] Sync: `tap-hubspot --config config.json --properties catalog.json` — verify sync runs normally with an access-filtered catalog
- [x] Unit Tests: `pytest tap_hubspot/tests/unittests/test_discovery.py` — all tests pass
- [x] Integration test: Running
 
# Risks
- Discovery now makes one extra API call per parent stream (~13 lightweight requests during `--discover`)
- Taps with fully restricted credentials will now fail fast at discovery instead of silently at sync time (intentional)
- `contact_lists` uses a POST probe; any change to that endpoint's auth behavior would affect the access check 
 
# Rollback steps
 - Revert this branch

#### AI generated code
https://internal.qlik.dev/general/ways-of-working/code-reviews/#guidelines-for-ai-generated-code
- [ ] this PR has been written with the help of GitHub Copilot or another generative AI tool

### DIFF
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -443,7 +443,8 @@ def _get_accessible_streams(streams):
 
     # If ALL parent streams are inaccessible, raise error
     parent_streams = [s for s in streams if not s.parent_tap_stream_id]
-    if inaccessible_streams and len(inaccessible_streams) == len(parent_streams):
+    inaccessible_parent_streams = [s for s in inaccessible_streams if not s.parent_tap_stream_id]
+    if inaccessible_parent_streams and len(inaccessible_parent_streams) == len(parent_streams):
         raise HubspotForbiddenError(
             "403 Forbidden: No read access to supported streams. Data collection cannot start."
         )
--- a/tap_hubspot/__init__.py
+++ b/tap_hubspot/__init__.py
@@ -35,6 +35,10 @@ class DependencyException(Exception):
 class UriTooLongException(Exception):
     pass
 
+class HubspotForbiddenError(Exception):
+    """Raised when the API returns 403 Forbidden for all stream endpoints."""
+    pass
+
 class DataFields:
     offset = 'offset'
 
@@ -66,6 +70,24 @@ CONFIG = {
     "hapikey": None,
     "include_inactives": None,
     "select_fields_by_default": None,
+}
+
+# Maps each parent stream to the endpoint key and minimal params for access probing.
+# Child streams are excluded since their access is governed by the parent.
+STREAM_ACCESS_ENDPOINTS = {
+    "subscription_changes": {"endpoint": "subscription_changes", "params": {"startTimestamp": 0, "endTimestamp": 1, "limit": 1}},
+    "email_events":         {"endpoint": "email_events", "params": {"startTimestamp": 0, "endTimestamp": 1, "limit": 1}},
+    "contacts":             {"endpoint": "contacts", "params": {"limit": 1}},
+    "deals":                {"endpoint": "deals_all", "params": {"limit": 1}},
+    "companies":            {"endpoint": "companies_all", "params": {"limit": 1}},
+    "tickets":              {"endpoint": "tickets", "params": {"limit": 1}},
+    "owners":               {"endpoint": "owners", "params": {"limit": 1}},
+    "forms":                {"endpoint": "forms"},
+    "workflows":            {"endpoint": "workflows"},
+    "contact_lists":        {"endpoint": "contact_lists", "method": "POST", "body": {"count": 1}},
+    "engagements":          {"endpoint": "engagements_all", "params": {"limit": 1}},
+    "campaigns":            {"endpoint": "campaigns_all", "params": {"limit": 1}},
+    "deal_pipelines":       {"endpoint": "deal_pipelines"},
 }
 
 ENDPOINTS = {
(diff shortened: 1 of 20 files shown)


tap_hubspot/tests/unittests/test_discovery.py
```diff
@@ -0,0 +1,264 @@
...
+        result = check_stream_access("contact_lists")
+        self.assertTrue(result)
+        mock_post.assert_called_once()
+
+    @patch('tap_hubspot.post_search_endpoint')
+    def test_contact_lists_returns_false_on_403(self, mock_post):
+        """contact_lists returning 403 via POST should return False."""
+        mock_post.side_effect = SourceUnavailableException("Forbidden")
+        result = check_stream_access("contact_lists")
+        self.assertFalse(result)
```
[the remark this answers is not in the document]
MuralidharT03: Addressed

tap_hubspot/__init__.py
```diff
@@ -347,6 +369,101 @@ def request(url, params=None):
...
+        return True
+    except SourceUnavailableException:
+        LOGGER.warning(
+            "Stream '%s' is not accessible with the provided credentials (403 Forbidden). "
+            "Excluding from catalog.",
+            stream_name,
+        )
+        return False
+    except requests.exceptions.HTTPError as exc:
+        if exc.response is not None and exc.response.status_code == 403:
```
RushiT0122:
```suggestion
        if exc.response and exc.response.status_code == 403:
```

RushiT0122: This will fail is `response={}` and `exc.response.status_code` would fail.

tap_hubspot/__init__.py
```diff
@@ -347,6 +369,101 @@ def request(url, params=None):
...
+            LOGGER.warning(
+                "Stream '%s' is not accessible with the provided credentials (403 Forbidden). "
+                "Excluding from catalog.",
+                stream_name,
+            )
+            return False
+        raise
```
RushiT0122, line 416:
```suggestion
    except SourceUnavailableException:
        return False
    except requests.exceptions.HTTPError as exc:
        if exc.response is not None and exc.response.status_code == 403:
            return False
        raise
```

tap_hubspot/__init__.py
```diff
@@ -347,6 +369,101 @@ def request(url, params=None):
...
+    for stream in streams:
+        # Child streams inherit access from their parent
+        if stream.parent_tap_stream_id:
+            accessible_endpoints.append(stream)
+            continue
+
+        if check_stream_access(stream.tap_stream_id):
+            accessible_endpoints.append(stream)
+        else:
+            inaccessible_endpoints.append(stream)
```
RushiT0122:
```suggestion
        else:
            LOGGER.warning(
               "Stream '%s' is not accessible with the provided credentials (403 Forbidden). "
               "Excluding from catalog.",
               stream.tap_stream_id,
            )
            inaccessible_endpoints.append(stream)
```

tap_hubspot/__init__.py
```diff
@@ -347,6 +369,101 @@ def request(url, params=None):
...
+
+    # Prune child streams whose parent was excluded
+    excluded_parent_ids = {s.tap_stream_id for s in inaccessible_endpoints}
+    pruned = []
```
RushiT0122: Please rename it properly.

tap_hubspot/__init__.py
```diff
@@ -347,6 +369,101 @@ def request(url, params=None):
...
+            pruned.append(stream)
+    accessible_endpoints = pruned
+
+    # If ALL parent streams are inaccessible, raise error
+    parent_streams = [s for s in streams if not s.parent_tap_stream_id]
+    if inaccessible_endpoints and len(inaccessible_endpoints) == len(parent_streams):
+        raise HubspotForbiddenError(
+            "HTTP-error-code: 403, Error: The account credentials supplied do not have 'read' access to any "
+            "of the streams supported by the tap. Data collection cannot be initiated due to lack of permissions."
+        )
```
RushiT0122, line 464:
```suggestion
        raise HubspotForbiddenError( "403 Forbidden: No read access to supported streams. Data collection cannot start.")
```

tap_hubspot/__init__.py
```diff
@@ -347,6 +369,101 @@ def request(url, params=None):
...
+
+    if inaccessible_endpoints:
+        LOGGER.warning(
+            "The account credentials supplied do not have 'read' access to the following stream(s): %s. "
+            "These streams have been excluded from the catalog.",
+            ", ".join(s.tap_stream_id for s in inaccessible_endpoints),
+        )
```
RushiT0122, line 470:
```suggestion
        LOGGER.warning( 
              "Excluding inaccessible streams: %s",
              ", ".join(s.tap_stream_id for s in inaccessible_endpoints),
       )
```

RushiT0122 (changes requested):

RushiT0122 (commented):
Minor fix needed, if parent is excluded then we need to add child stream in inaccessible list as well.
e.g.

```
WARNING Excluding inaccessible streams: contact_lists, list_memberships
```

RushiT0122 (approved):

RushiT0122 (changes requested):
Instead of excluding inaccessible streams, we should use to validate expected behavior. For that you can add property like method like below and define method `expected_inaccessible_streams()`

```
            "workflows": {
                self.PRIMARY_KEYS: {"id"},
                self.REPLICATION_METHOD: self.INCREMENTAL,
                self.REPLICATION_KEYS: {"updatedAt"},
                self.FORBIDDEN_STREAM: True,
                self.OBEYS_START_DATE: True
            },
```

RushiT0122 (changes requested):
Remove `workflows` stream from the stream exclusion list from integration tests.

RushiT0122 (changes requested):
**Findings**

**[Medium]** __init__.py:363-382: API 401 responses are retried by request() and eventually converted by on_giveup() into a generic exception containing only the URL and params. Discovery does not fail fast with an authentication/configuration-specific error, and the original 401 reason is lost. Handle 401 explicitly and raise InvalidAuthException or another actionable auth error without probing every stream.

**[Low]** __init__.py:404-415: check_stream_access() reduces a 403 to False, and _get_accessible_streams() emits only a generic “403 Forbidden” warning. The API response body/reason is discarded, so users cannot determine which permission or scope is missing. Preserve a sanitized status/reason in the stream warning while ensuring credentials and authorization headers are excluded.

